PR #10, opened 2022-02-22
Yubrew/5-enforce-start-time

### LINKED ISSUE #5: Enforce start times in contracts

yubrew:
fix #5 

enforces start time at genesis mint or later `3/1/2022 00:00:00 GMT`
also enforces genesis mint or later when updating start time

### DIFF
--- a/contracts/minter/src/contract_tests.rs
+++ b/contracts/minter/src/contract_tests.rs
@@ -859,14 +859,7 @@ fn test_start_time_before_genesis() {
         },
     };
     let minter_addr = router
-        .instantiate_contract(
-            minter_code_id,
-            creator.clone(),
-            &msg,
-            &creation_fee,
-            "Minter",
-            None,
-        )
+        .instantiate_contract(minter_code_id, creator, &msg, &creation_fee, "Minter", None)
         .unwrap();
 
     let res: StartTimeResponse = router
--- a/contracts/minter/src/contract_tests.rs
+++ b/contracts/minter/src/contract_tests.rs
@@ -6,7 +6,7 @@ use cw_multi_test::{App, BankSudo, Contract, ContractWrapper, Executor, SudoMsg}
 use cw_utils::Expiration;
 use sg721::msg::InstantiateMsg as Sg721InstantiateMsg;
 use sg721::state::{Config, RoyaltyInfo};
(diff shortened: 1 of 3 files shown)


packages/sg-std/src/lib.rs
```diff
@@ -2,6 +2,8 @@ mod msg;
 mod route;
 
 pub const NATIVE_DENOM: &str = "ustars";
+// 3/1/2022 00:00:00GMT
+pub const GENESIS_MINT_START_TIME: u64 = 1646092800000000000;
```
jhernandezb, line 6: Do we want this to be a constant? Why not as part of the instantiation ?  Otherwise future mints will be stuck to this date (allowing front running)

jhernandezb: ooh nvm I just read the whole thing, thanks to github diff for being unhelpful

jhernandezb: on the other side should we check is not a time in the past (start_time> env.block_time) ?

contracts/minter/src/contract_tests.rs
```diff
@@ -293,10 +301,22 @@ fn initialization() {
 #[test]
 fn happy_path() {
     let mut router = mock_app();
+    let _timestamp = setup_block_time(&mut router, GENESIS_MINT_START_TIME + 1).unwrap();
```
shanev, line 304: Can just remove `_timestamp` right? No big deal though. Developer decision. It's a test :)

shanev (approved):
utACK :)

shanev:
Better to use a branch name like `yubrew/5-enforce-start-time`. Git has issues with branch names that have more than one level of nesting with `/`.

packages/sg-std/src/lib.rs
yubrew: After genesis mint if the start_date is not set, people can mint right away. Don't want to bloat the contracts with too much logic either.

I'm not sure what the right approach is

contracts/minter/src/contract_tests.rs
yubrew: yep, cleaned it up!

JakeHartnell (approved):

PR #21, opened 2022-02-24
Standard fee handler

shanev:


packages/sg-std/src/fees.rs
```diff
@@ -0,0 +1,49 @@
...
+    let burn_coin = coin(burn_fee.u128(), NATIVE_DENOM);
+    // send fee to contract to be burned
+    let send_fee_msg = BankMsg::Send {
+        to_address: env.contract.address.to_string(),
+        amount: vec![burn_coin.clone()],
+    };
+    // burn half the fee
+    let fee_burn_msg = BankMsg::Burn {
+        amount: vec![burn_coin],
+    };
```
JakeHartnell, line 30: So we're sending it to a contract to be burned and burning it?

JakeHartnell: Is one of these not needed?

shanev: Yeah exactly. Know of a better way? Another way is to send it to an address where the private key is destroyed.

JakeHartnell (approved):

PR #26, opened 2022-02-24
Upgrading to cw-plus v0.12.1

shanev:


contracts/minter/Cargo.toml
```diff
@@ -40,12 +40,12 @@ optimize = """docker run --rm -v "$(pwd)":/code \
...
-cosmwasm-storage = { version = "1.0.0-beta" }
-cw-storage-plus = "0.11"
-cw2 = "0.11"
-cw721 = "0.11"
-cw721-base = { version = "0.11", features = ["library"] }
+cosmwasm-std = { version = "1.0.0-beta5" }
+cosmwasm-storage = { version = "1.0.0-beta5" }
+cw-storage-plus = "0.12.1"
+cw2 = "0.12.1"
+cw721 = "0.11.0"
```
JakeHartnell, line 47: Latest is `0.11.1`.

You might already know this, but the end can be left off. For example, `1.0.0-beta` or `0.11`. Unless there is an `=` sign, Rust will bump dependencies similar to using the `^` in Node.js. Note, that the `Cargo.lock` file didn't change much even after all these modifications we're made.

https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio

shanev: `cargo add cw721` is not picking up v0.11.1. Looks like the crate hasn't been updated on crates.io: https://crates.io/crates/cw721.

Well aware that you can leave out the end and let Cargo pick up the latest version. But this is bad practice, especially for contracts. Version numbers should be super explicit for financial code.

jhernandezb (approved):

shanev:
Will upgrade to cw721 v0.11.1 in separate PR after the crate has been pushed.

yubrew (approved):
👍

PR #63, opened 2022-02-26
Add price to whitelist

shanev:
Fixes #63


shanev:
Too far deviated from `main`. Replaced with a fresh #70.

PR #66, opened 2022-02-26
Yubrew/patch-wl-price

yubrew:


contracts/minter/src/contract.rs
```diff
@@ -332,8 +333,29 @@ pub fn execute_batch_mint(
         return Err(ContractError::MaxBatchMintLimitExceeded {});
     }
 
+    // calculate the mint fee
+    // if in whitelist, use whitelist price. else use mint price
+    let mint_price: Coin = if let Some(whitelist) = config.whitelist {
```
shanev, line 338: Can break into a function?

shanev (approved):
Can probably break out common code into a function.

JakeHartnell (approved):

contracts/minter/src/contract.rs
```diff
@@ -377,27 +399,48 @@ fn _execute_mint(
...
+            let unit_price: UnitPriceResponse = deps
+                .querier
+                .query_wasm_smart(whitelist, &WhitelistQueryMsg::UnitPrice {})?;
+            unit_price.unit_price
+        } else {
+            config.unit_price.clone()
+        }
+    } else {
+        config.unit_price.clone()
+    };
```
jhernandezb, line 420: can we refactor into a method  seems like its the same code for batch and regular mint.

shanev:
Sorry I merged this too quickly. Yes can refactor.

contracts/minter/src/contract.rs
yubrew: yes i was thinking the same!

PR #74, opened 2022-02-27
Yubrew/67 creators have free mints for mint_to and mint_for

### LINKED ISSUE #67: Mint to from creator should be free for airdrops
Basically admins should mint for free.

yubrew:
fix #67 
`creator` can mint for free using `mint_for` and `mint_to`

`_execute_mint` has an extra param `admin_no_fee: true` passed by `mint_for` and `mint_to`

this skips `community_fund`, `burn`, and `seller` messages because price is 0

### DIFF
--- a/contracts/minter/src/contract.rs
+++ b/contracts/minter/src/contract.rs
@@ -40,7 +40,6 @@ const STARTING_BATCH_MINT_LIMIT: u64 = 5;
 const STARTING_PER_ADDRESS_LIMIT: u64 = 5;
 const MIN_MINT_PRICE: u128 = 50_000_000;
 const MINT_FEE_PERCENT: u64 = 10;
-const ADMIN_MINT_PRICE: u128 = 0;
 
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn instantiate(
@@ -277,7 +276,7 @@ pub fn execute_mint_sender(
         }
     }
 
-    _execute_mint(deps, env, info, action, None, None)
+    _execute_mint(deps, env, info, action, false, None, None)
 }
 
 pub fn execute_mint_to(
@@ -296,7 +295,7 @@ pub fn execute_mint_to(
         ));
     }
 
-    _execute_mint(deps, env, info, action, Some(recipient), None)
+    _execute_mint(deps, env, info, action, true, Some(recipient), None)
 }
 
 pub fn execute_mint_for(
@@ -316,13 +315,21 @@ pub fn execute_mint_for(
         ));
     }
 
-    _execute_mint(deps, env, info, action, Some(recipient), Some(token_id))
+    _execute_mint(
+        deps,
+        env,
+        info,
+        action,
+        true,
+        Some(recipient),
+        Some(token_id),
+    )
 }
 
 pub fn execute_batch_mint(
     deps: DepsMut,
     env: Env,
-    info: MessageInfo,
+    _info: MessageInfo,
     num_mints: u64,
 ) -> Result<Response, ContractError> {
     let config = CONFIG.load(deps.storage)?;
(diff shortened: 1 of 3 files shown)


contracts/minter/src/contract_tests.rs
```diff
@@ -371,15 +381,18 @@ fn happy_path() {
         .unwrap();
     assert_eq!(res.owner, buyer.to_string());
 
-    // Buyer can't call MintTo
+    // // Buyer can't call MintTo
```
shanev, line 384:
```suggestion
```
Guessing this comment can be deleted?

contracts/minter/src/contract.rs
```diff
@@ -440,17 +455,21 @@ fn _execute_mint(
...
+            amount: vec![coin(seller_amount.u128(), config.unit_price.denom)],
+        }]
     };
-    msgs.append(&mut vec![seller_msg.into()]);
 
     Ok(Response::default()
         .add_attribute("action", action)
         .add_messages(msgs)
-        .add_messages(network_fee_msg))
+        .add_messages(network_fee_msgs)
```
shanev, line 471: So instead of sending empty messages here, another approach is to create a mutable messages array, and just append the relevant messages to it.

contracts/minter/src/contract.rs
```diff
@@ -524,10 +543,17 @@ pub fn execute_update_batch_mint_limit(
...
 
-pub fn mint_price(deps: Deps) -> Result<Coin, StdError> {
-    // if in whitelist, use whitelist price. else use config unit price
+pub fn mint_price(deps: Deps, admin_no_fee: bool) -> Result<Coin, StdError> {
+    // if admin => admin mint fee,
+    // else if in whitelist => whitelist price
+    // else => config unit price
     let config = CONFIG.load(deps.storage)?;
-    let mint_price: Coin = if let Some(whitelist) = config.whitelist {
+    let mint_price: Coin = if admin_no_fee {
```
shanev, line 551: Instead of this, you can also make the `mint_price` an `Option` type right?

shanev (changes requested):
Thanks for getting this working! Going to hack on it a bit, I think there's some refactoring we can do.

packages/sg-std/src/fees.rs
```diff
@@ -12,19 +12,17 @@ pub fn burn_and_distribute_fee(
     info: &MessageInfo,
     fee_amount: u128,
 ) -> Result<Vec<SubMsg>, FeeError> {
-    let payment = must_pay(info, NATIVE_DENOM)?;
+    let payment = may_pay(info, NATIVE_DENOM)?;
```
JakeHartnell, line 15: Not sure why this switched because you only call this function if there is a fee...

contracts/minter/src/contract.rs
yubrew: i like it

contracts/minter/src/contract.rs
shanev: i.e: The function signature can probably just be `fn mint_price(deps: Deps, admin_no_fee: bool) -> Option<Coin>`

packages/sg-std/src/fees.rs
yubrew: good catch, fixed

contracts/minter/src/contract.rs
```diff
@@ -524,10 +542,17 @@ pub fn execute_update_batch_mint_limit(
     Ok(Response::new().add_attribute("action", "update_batch_mint_limit"))
 }
 
-pub fn mint_price(deps: Deps) -> Result<Coin, StdError> {
-    // if in whitelist, use whitelist price. else use config unit price
+pub fn mint_price(deps: Deps, admin_no_fee: bool) -> Result<Coin, StdError> {
```
shanev, line 545: Does this function need to be public?

shanev, line 545:
```suggestion
fn mint_price(deps: Deps, admin_no_fee: bool) -> Result<Coin, StdError> {
```

shanev (approved):
🤘

PR #78, opened 2022-02-28
Yubrew/77-query-mint-price

### LINKED ISSUE #77: query mint price
query mint price so front end can access correct mint price as `Coin`

yubrew:
fix #77

query access to mint_price, probably useful for front end

### DIFF
--- a/contracts/minter/src/contract.rs
+++ b/contracts/minter/src/contract.rs
@@ -13,8 +13,8 @@ use url::Url;
 
 use crate::error::ContractError;
 use crate::msg::{
-    ConfigResponse, ExecuteMsg, InstantiateMsg, MintableNumTokensResponse, QueryMsg,
-    StartTimeResponse,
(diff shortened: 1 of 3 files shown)


contracts/minter/src/contract.rs
```diff
@@ -621,6 +622,12 @@ fn query_mintable_num_tokens(deps: Deps) -> StdResult<MintableNumTokensResponse>
         count: count as u64,
     })
 }
+
+fn query_mint_price(deps: Deps) -> StdResult<MintPriceResponse> {
+    let mint_price = mint_price(deps, false)?;
+    Ok(MintPriceResponse { price: mint_price })
```
jhernandezb, line 628: should we add whitelist price if available?

yubrew: mint_price gives whitelist price if whitelist is active, otherwise gives mint price

shanev (changes requested):
The frontend would need both the whitelist price and the public sale price separately

jhernandezb (approved):

shanev (approved):
Excellent solution!

PR #96, opened 2022-03-01
Better collection-level metadata

shanev:
Been having lots of confusion with creators and testers using the tooling about the redundant data in cw721 collection info and contract URI. This is an attempt to clear things up and and have a single source of truth for collection data. It basically moves everything we need from contract URI into sg721.

This introduces two required fields, `image` that serves as a preview image for the entire collection, and `description`. `image` doesn't need to be an IPFS link since it's more marketplace-level metadata and is not a traded asset like the individual NFTs. Note this does not break compatibility with cw721 since instantiation is not part of the cw721 spec.

This also introduces an optional `external_link` that can be a website, etc. 

These two fields are all we needed from contract URI anyway. The name, description, seller_fee (royalties) were redundant because they're already part of cw721 `ContractInfo` and sg721 royalties. Contract URI is used on OpenSea mainly due to a limitation of ERC721 that doesn't include basic metadata like name and description.

Also removed `creator` since it's already part of `ContractInfoResponse`, and `payment_address` is already part of royalties.

I think this cleans and simplifies a lot, and makes collection info more explicitly defined. Also it's one less thing for creators to worry about and less error prone. Also collection info should be easily queried and can easily live on-chain. No need to hide behind an IPFS link. It's also not part of ICS-721 so there's no benefit in interoperability either.

### DIFF
--- a/contracts/minter/schema/instantiate_msg.json
+++ b/contracts/minter/schema/instantiate_msg.json
@@ -74,24 +74,20 @@
         }
       }
     },
-    "Config": {
+    "CollectionInfo": {
       "type": "object",
+      "required": [
+        "image"
+      ],
       "properties": {
-        "contract_uri": {
+        "external_link": {
           "type": [
             "string",
             "null"
           ]
         },
-        "creator": {
-          "anyOf": [
-            {
-              "$ref": "#/definitions/Addr"
-            },
-            {
-              "type": "null"
-            }
-          ]
+        "image": {
+          "type": "string"
         },
         "royalties": {
           "anyOf": [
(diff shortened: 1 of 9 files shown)


yubrew:
This is a great improvement, a source of frustration for me when I tried to get the test nft collection up.

shanev:
@yubrew Good points. I think it makes sense to make all of `CollectionInfo` editable. The ones that should be static are `name` and `symbol` which are part of cw721 and non-editable.

shanev:
Added https://github.com/public-awesome/stargaze-contracts/issues/99 to the v2 board. Want to keep changes minimal before upload.

yubrew (approved):

jhernandezb (approved):

PR #97, opened 2022-03-01
DO NOT MERGE: Repo review

orkunkl:
I will use this PR to leave general review comments. Please **Do Not Merge**


contracts/minter/src/contract.rs
```diff
@@ -0,0 +1,602 @@
...
+    QueryMsg, StartTimeResponse,
+};
+use crate::state::{Config, CONFIG, MINTABLE_TOKEN_IDS, SG721_ADDRESS};
+use sg_std::{burn_and_distribute_fee, StargazeMsgWrapper, GENESIS_MINT_START_TIME, NATIVE_DENOM};
+use whitelist::msg::{
+    ConfigResponse as WhitelistConfigResponse, HasMemberResponse, QueryMsg as WhitelistQueryMsg,
+    UnitPriceResponse,
+};
+
+pub type Response = cosmwasm_std::Response<StargazeMsgWrapper>;
```
orkunkl, line 26: This could be used by many smart contracts. Maybe we can pull this to a package?

contracts/minter/src/contract.rs
```diff
@@ -0,0 +1,602 @@
...
+        });
+    }
+
+    // Initially set per_address_limit if no msg
+    let per_address_limit: Option<u32> = msg.per_address_limit.or(Some(STARTING_PER_ADDRESS_LIMIT));
+
+    let whitelist_addr: Option<Addr> = match msg.whitelist {
+        Some(wl) => Some(deps.api.addr_validate(&wl)?),
+        None => None,
+    };
```
orkunkl, line 92:
```suggestion
    let whitelist = msg
        .whitelist.and_then(|w| deps.api.addr_validate(w.as_str()).ok());
```

contracts/minter/src/contract.rs
```diff
@@ -0,0 +1,602 @@
...
+    let start_time = match msg.start_time {
+        Some(st) => {
+            if st < default_start_time {
+                default_start_time
+            } else {
+                st
+            }
+        }
+        None => default_start_time,
+    };
```
orkunkl, line 105:
``` 
    let start_time = msg.start_time.map_or(default_start_time, |st|  max(st, default_start_time));
```

contracts/minter/src/contract.rs
```diff
@@ -0,0 +1,602 @@
...
+) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    let sg721_address = SG721_ADDRESS.load(deps.storage)?;
+    let action = "mint_sender";
+    let mut pub_mint: bool = false;
+
+    // check if a whitelist exists and not ended
+    // sender has to be whitelisted to mint
+    if let Some(whitelist) = config.whitelist {
+        let wl_config: WhitelistConfigResponse = deps
```
orkunkl, line 210: You will need to run these queries a lot. Same for external contracts. 
Best to move these under a helpers in a package and use it https://github.com/CosmWasm/cw-plus/blob/main/packages/cw4/src/helpers.rs

contracts/minter/src/contract.rs
shanev: Good idea, but it depends on a custom SDK module.

contracts/minter/src/contract.rs
```diff
@@ -0,0 +1,602 @@
...
+    info: MessageInfo,
+) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    let sg721_address = SG721_ADDRESS.load(deps.storage)?;
+    let action = "mint_sender";
+    let mut pub_mint: bool = false;
+
+    // check if a whitelist exists and not ended
+    // sender has to be whitelisted to mint
+    if let Some(whitelist) = config.whitelist {
```
orkunkl, line 209: You can move branched logics under new functions, this will increase readability immensely

contracts/minter/src/contract.rs
```diff
@@ -0,0 +1,602 @@
...
+}
+
+pub fn execute_mint_to(
+    deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    recipient: Addr,
+) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+    let action = "mint_to";
```
orkunkl, line 280: Interesting move :D

shanev:
Thanks for the comments @orkunkl, think you can rebase this branch off of `main`?

contracts/minter/src/contract.rs
```diff
@@ -0,0 +1,602 @@
...
+        sg721_code_id: msg.sg721_code_id,
+        unit_price: msg.unit_price,
+        per_address_limit,
+        whitelist: whitelist_addr,
+        start_time: Some(start_time),
+    };
+    CONFIG.save(deps.storage, &config)?;
+
+    // save mintable token ids map
+    for token_id in 1..=msg.num_tokens {
```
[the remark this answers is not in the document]
shanev: I actually think the range operator is more readable here. Any reason `repeat` is more preferable?

contracts/minter/src/contract.rs
shanev: This is a good idea. An optimization for the future for sure.

contracts/minter/src/contract.rs
shanev: Yeah this whole block can be refactored, but it works!

contracts/minter/src/contract.rs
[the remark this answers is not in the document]
orkunkl: I tried some stuff, just to use more functional. but yo are right, this do not make sense

contracts/minter/src/msg.rs
```diff
@@ -0,0 +1,68 @@
...
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum ExecuteMsg {
+    Mint {},
+    SetWhitelist { whitelist: String },
+    UpdateStartTime(Expiration),
+    UpdatePerAddressLimit { per_address_limit: u32 },
+    MintTo { recipient: Addr },
+    MintFor { token_id: u64, recipient: Addr },
```
orkunkl, line 28: You should not accept Addr in messages. `Addr` is a validated address. Use string and validate inside execute

contracts/minter/src/state.rs
```diff
@@ -0,0 +1,22 @@
...
+    pub sg721_code_id: u64,
+    pub unit_price: Coin,
+    pub whitelist: Option<Addr>,
+    pub start_time: Option<Expiration>,
+    pub per_address_limit: Option<u32>,
+}
+
+pub const CONFIG: Item<Config> = Item::new("config");
+pub const SG721_ADDRESS: Item<Addr> = Item::new("sg721_address");
+pub const MINTABLE_TOKEN_IDS: Map<u64, Empty> = Map::new("mt");
```
orkunkl, line 22: Instead of removing data, you can set the value as bool and flag it true. In fact `Empty` uses more space than `bool`

contracts/minter/src/contract.rs
shanev: I don't think `max` ordinal is implemented for `Expiration`. Will check (or contribute if not).

contracts/minter/src/msg.rs
```diff
@@ -0,0 +1,68 @@
...
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+pub struct ConfigResponse {
+    pub admin: Addr,
+    pub base_token_uri: String,
+    pub num_tokens: u64,
+    pub per_address_limit: Option<u32>,
+    pub sg721_address: Addr,
+    pub sg721_code_id: u64,
+    pub start_time: Option<Expiration>,
```
orkunkl, line 48: `Expiration` struct could have `Never` value which you should check. We introduced a new struct called `Scheduled` for this case.

contracts/whitelist/src/contract.rs
```diff
@@ -0,0 +1,450 @@
...
+    }
+
+    config.start_time = start_time;
+    CONFIG.save(deps.storage, &config)?;
+    Ok(Response::new()
+        .add_attribute("action", "update_start_time")
+        .add_attribute("start_time", start_time.to_string()))
+}
+
+pub fn execute_update_end_time(
```
orkunkl, line 141: You can use generalized `execute_update_config` function with optional values like in https://github.com/deus-labs/junotools-contracts/blob/main/contracts/jt-airdrop-controller/src/contract.rs#L91.
Makes the code and msgs smaller. This is optional.

contracts/royalty-group/src/contract.rs
```diff
@@ -0,0 +1,601 @@
...
+    TOTAL.save(deps.storage, &total)?;
+    Ok(MemberChangedHookMsg { diffs })
+}
+
+pub fn execute_distribute_funds(
+    deps: DepsMut,
+    info: MessageInfo,
+    funds: Coin,
+) -> Result<Response, ContractError> {
+    let messages: StdResult<Vec<_>> = MEMBERS
```
orkunkl, line 161: This could get quite expensive, you can use an optimized reward distribution algorithm/contract based on the logic here: https://github.com/CosmWasm/cw-tokens/pull/8

contracts/minter/src/msg.rs
```diff
@@ -0,0 +1,68 @@
...
+}
+
+#[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
+#[serde(rename_all = "snake_case")]
+pub enum ExecuteMsg {
+    Mint {},
+    SetWhitelist { whitelist: String },
+    UpdateStartTime(Expiration),
+    UpdatePerAddressLimit { per_address_limit: u32 },
+    MintTo { recipient: String },
```
orkunkl, line 27: I think you should add a /// comment to differentiate `MintTo` and `MintFor`. Easy to understand without looking at code.

PR #108, opened 2022-03-02
refactor whitelist

yubrew:


contracts/minter/src/contract.rs
```diff
@@ -522,13 +508,19 @@ pub fn mint_price(deps: Deps, admin_no_fee: bool) -> Result<Coin, StdError> {
...
+}
+
 #[cfg_attr(not(feature = "library"), entry_point)]
 pub fn query(deps: Deps, _env: Env, msg: QueryMsg) -> StdResult<Binary> {
     match msg {
         QueryMsg::Config {} => to_binary(&query_config(deps)?),
         QueryMsg::StartTime {} => to_binary(&query_start_time(deps)?),
         QueryMsg::MintableNumTokens {} => to_binary(&query_mintable_num_tokens(deps)?),
         QueryMsg::MintPrice {} => to_binary(&query_mint_price(deps)?),
+        QueryMsg::MinterAddresses {} => to_binary(&query_minter_addresses(deps)?),
```
shanev, line 523: Would this ever need be queried? Isn't it more for internal accounting?

shanev (changes requested):

contracts/minter/src/contract.rs
yubrew: Initially thinking creators might want to know distribution of addresses doing the minting. Now that I'm thinking about it, seems like something no one asked for and few people really want. Removing this bloat.

JakeHartnell:
Needs `cargo schema` update.

shanev (approved):
LGTM

yubrew:
updated schema

jhernandezb:
Can we add a MintCount query? this will make the airdrop claimer easier

JakeHartnell (approved):
+1 on @jhernandezb's request but can be a separate PR IMO.

shanev:
Created #110

PR #114, opened 2022-03-03
Idiomatic royalty validation

shanev:
We weren't using `String` for the royalty payment address. This fixes it to be more consistent with the rest of the contracts, and also introduces a separate message type for royalty responses that uses strings instead of `Addr`.

### DIFF
--- a/contracts/sg721/src/state.rs
+++ b/contracts/sg721/src/state.rs
@@ -1,15 +1,14 @@
-use crate::ContractError;
 use cosmwasm_std::{Addr, Decimal};
 use cw_storage_plus::Item;
 use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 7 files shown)


contracts/sg721/src/state.rs
```diff
@@ -1,15 +1,14 @@
...
 use serde::{Deserialize, Serialize};
 
 #[derive(Serialize, Deserialize, Clone, Debug, PartialEq, JsonSchema)]
-pub struct CollectionInfo {
+pub struct CollectionInfo<T> {
     pub description: String,
     pub image: String,
     pub external_link: Option<String>,
-    pub royalties: Option<RoyaltyInfo>,
+    pub royalty_info: Option<T>,
```
JakeHartnell, line 11: This is _really_ nice. Will provide a lot more flexibility in the future. :heart:

JakeHartnell (approved):

PR #119, opened 2022-03-04
Whitelist refactor

### LINKED ISSUE #106: refactor whitelist

### LINKED ISSUE #107: delete `remove` function from whitelist

shanev:
Fixes #106. Fixes #107.

### DIFF
--- a/contracts/minter/src/contract.rs
+++ b/contracts/minter/src/contract.rs
@@ -20,8 +20,7 @@ use crate::state::{
 };
 use sg_std::{burn_and_distribute_fee, StargazeMsgWrapper, GENESIS_MINT_START_TIME, NATIVE_DENOM};
 use whitelist::msg::{
-    ConfigResponse as WhitelistConfigResponse, HasMemberResponse, IsActiveResponse,
-    QueryMsg as WhitelistQueryMsg,
+    ConfigResponse as WhitelistConfigResponse, HasMemberResponse, QueryMsg as WhitelistQueryMsg,
 };
 
 pub type Response = cosmwasm_std::Response<StargazeMsgWrapper>;
@@ -202,11 +201,7 @@ pub fn execute_mint_sender(
             .querier
             .query_wasm_smart(whitelist.clone(), &WhitelistQueryMsg::Config {})?;
 
-        let is_active_res: IsActiveResponse = deps
-            .querier
-            .query_wasm_smart(whitelist.clone(), &WhitelistQueryMsg::IsActive {})?;
-
-        if is_active_res.is_active {
+        if wl_config.is_active {
             let res: HasMemberResponse = deps.querier.query_wasm_smart(
                 whitelist,
                 &WhitelistQueryMsg::HasMember {
@@ -487,13 +482,9 @@ pub fn mint_price(deps: Deps, admin_no_fee: bool) -> Result<Coin, StdError> {
     } else if let Some(whitelist) = config.whitelist {
         let wl_config: WhitelistConfigResponse = deps
             .querier
-            .query_wasm_smart(whitelist.clone(), &WhitelistQueryMsg::Config {})?;
+            .query_wasm_smart(whitelist, &WhitelistQueryMsg::Config {})?;
 
-        let is_active_res: IsActiveResponse = deps
-            .querier
-            .query_wasm_smart(whitelist, &WhitelistQueryMsg::IsActive {})?;
-
-        if is_active_res.is_active {
+        if wl_config.is_active {
             wl_config.unit_price
         } else {
             config.unit_price.clone()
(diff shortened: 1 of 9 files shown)


contracts/whitelist/src/contract.rs
```diff
@@ -157,35 +178,32 @@ pub fn execute_update_end_time(
...
         if config.num_members >= MAX_MEMBERS {
             return Err(ContractError::MembersExceeded {
                 expected: MAX_MEMBERS,
                 actual: config.num_members,
             });
         }
         let addr = deps.api.addr_validate(&add)?;
+        if WHITELIST.has(deps.storage, addr.clone()) {
+            return Err(ContractError::DuplicateMember(addr.to_string()));
+        }
```
jhernandezb, line 202: we could also do
```
if !WHITELIST.has(deps.storage, addr.clone()) {
        WHITELIST.save(deps.storage, addr, &true)?;
        config.num_members += 1;
}
```
but maybe better to be explicit about the issue

contracts/whitelist/src/contract.rs
```diff
@@ -157,35 +178,32 @@ pub fn execute_update_end_time(
...
         WHITELIST.save(deps.storage, addr, &true)?;
         config.num_members += 1;
     }
 
-    for remove in msg.remove.into_iter() {
-        let addr = deps.api.addr_validate(&remove)?;
-        WHITELIST.remove(deps.storage, addr);
-        config.num_members -= 1;
-    }
-
```
jhernandezb, line 188: not totally sure if we should remove this, what if someone actually messed up and needs to delete a single account? They will need to recreate an entire whitelist just because a single mistake.  However maybe allow removing only if the whitelist time hasn't started.

contracts/minter/src/contract.rs
```diff
@@ -202,7 +202,11 @@ pub fn execute_mint_sender(
             .querier
             .query_wasm_smart(whitelist.clone(), &WhitelistQueryMsg::Config {})?;
 
-        if wl_config.is_active {
+        let is_active_res: IsActiveResponse = deps
+            .querier
+            .query_wasm_smart(whitelist.clone(), &WhitelistQueryMsg::IsActive {})?;
```
jhernandezb, line 207: do we really need an extra query? maybe a helper instead? We already have the wl_config above.

contracts/minter/src/contract.rs
```diff
@@ -483,8 +487,13 @@ pub fn mint_price(deps: Deps, admin_no_fee: bool) -> Result<Coin, StdError> {
     } else if let Some(whitelist) = config.whitelist {
         let wl_config: WhitelistConfigResponse = deps
             .querier
-            .query_wasm_smart(whitelist, &WhitelistQueryMsg::Config {})?;
-        if wl_config.is_active {
+            .query_wasm_smart(whitelist.clone(), &WhitelistQueryMsg::Config {})?;
+
+        let is_active_res: IsActiveResponse = deps
+            .querier
+            .query_wasm_smart(whitelist, &WhitelistQueryMsg::IsActive {})?;
```
jhernandezb, line 494: same here about an extra query.

jhernandezb (changes requested):

contracts/whitelist/src/contract.rs
shanev: This is how I had it originally, but decided to be explicit in case they made any errors and uploading a lot of duplicates.

contracts/whitelist/src/contract.rs
shanev: I felt that it over-complicates things. But you have a point.

contracts/minter/src/contract.rs
shanev: This is how it was before. It felt a bit strange to have a helper / computed value inside something that usually just returns data. But the gas savings might be worth it.

contracts/whitelist/src/contract.rs
shanev: A separate message makes a lot more sense and is less error prone.

contracts/whitelist/src/contract.rs
```diff
@@ -113,7 +116,7 @@ pub fn execute(
     match msg {
         ExecuteMsg::UpdateStartTime(time) => execute_update_start_time(deps, env, info, time),
         ExecuteMsg::UpdateEndTime(time) => execute_update_end_time(deps, env, info, time),
-        ExecuteMsg::UpdateMembers(msg) => execute_update_members(deps, env, info, msg),
+        ExecuteMsg::UpdateMembers(msg) => execute_add_members(deps, env, info, msg),
```
yubrew, line 119: change to `AddMembers`?

yubrew (approved):
just 1 comment to change `UpdateMembers` to `AddMembers`

contracts/whitelist/src/contract.rs
shanev: Added `AddMembers` and `RemoveMembers`

JakeHartnell (approved):

jhernandezb (approved):

PR #126, opened 2022-03-04
add member pricing and limit

### LINKED ISSUE #111: change whitelist pricing
whitelist fee should increase with increasing number of members.

- either up front in instantiation (this is probably easier and more user friendly)
- as you add and pass thresholds

yubrew:
Member limit can increase as a project grows in popularity. Member limit increases the whitelist creation fee. A creator has the option to increase the member limit, which may incur an upgrade fee.

Currently set at 100 stars per 1,000 members. This fee can be configurable by governance in the future.

There is still a `MAX_MEMBERS` upper limit.

fix #111


contracts/whitelist/src/contract.rs
```diff
@@ -284,6 +309,45 @@ pub fn execute_update_per_address_limit(
...
+
+    // if new limit crosses 1,000 members, requires upgrade fee. Otherwise, free upgrade.
+    let old_limit = Decimal::new(config.member_limit.into(), 3).ceil();
+    let new_limit = Decimal::new(member_limit.into(), 3).ceil();
+    let upgrade_fee: u128 = if new_limit > old_limit {
+        (new_limit - old_limit).to_u128().unwrap() * PRICE_PER_1000_MEMBERS
+    } else {
+        0
+    };
+    let payment = may_pay(&info, NATIVE_DENOM)?;
```
shanev, line 336: This doesn't appear to be actually charging the fee by calling `burn_and_distribute()`.

shanev (changes requested):
Nice implementation! Doesn't look like Fair Burn™️ is happening on member limit updates.

contracts/whitelist/src/contract.rs
yubrew: good catch!

shanev (approved):

jhernandezb (approved):

PR #138, opened 2022-03-06
Check WL start time

### LINKED ISSUE #136: execute_set_whitelist should check if contract already started
If a creator has a whitelist already set and already started, creator can still switch it.

shanev:
Fix #136


JakeHartnell (approved):

yubrew (changes requested):
Code looks good. Don't see a test for this error though?

PR #142, opened 2022-03-06
Expiration -> Timestamp

shanev:


yubrew (approved):
👍

contracts/whitelist/src/contract.rs
```diff
@@ -105,24 +108,21 @@ pub fn instantiate(
...
-    let genesis_start_time = Expiration::AtTime(Timestamp::from_nanos(GENESIS_MINT_START_TIME));
+    let genesis_start_time = Timestamp::from_nanos(GENESIS_MINT_START_TIME);
     if msg.start_time < genesis_start_time {
         return Err(ContractError::InvalidStartTime(
-            msg.start_time,
-            genesis_start_time,
+            start_time,
+            Expiration::AtTime(genesis_start_time),
         ));
     }
```
jhernandezb, line 127: a bit confused of the mix in using expiration and timestamps, maybe adding comments on which one is being used for new readers of the code.

contracts/minter/src/contract.rs
```diff
@@ -432,14 +430,14 @@ pub fn execute_update_start_time(
         return Err(ContractError::AlreadyStarted {});
     }
 
-    let genesis_start_time = Expiration::AtTime(Timestamp::from_nanos(GENESIS_MINT_START_TIME));
+    let genesis_start_time = Timestamp::from_nanos(GENESIS_MINT_START_TIME);
     let start_time = if start_time < genesis_start_time {
         genesis_start_time
     } else {
         start_time
```
jhernandezb, line 437: we are not checking that the new `start_time > env.block` time right ?

jhernandezb (changes requested):

contracts/minter/src/contract.rs
jhernandezb: also maybe for updating doesn't makes sense to default to genesis time, just in instantiation. You have to be explicit about the time you want to update to.

shanev: Also fixed this.

jhernandezb (approved):

PR #143, opened 2022-03-06
Some refactorings

shanev:
* Changed `num_tokens` to a `u32` because I don't think there's a reason for it to be a `u64` considering the max collection size.
* Flattened some code


yubrew (approved):
👍

jhernandezb (approved):

PR #154, opened 2022-03-07
Already set whitelist check

shanev:


contracts/minter/src/contract.rs
```diff
@@ -191,6 +191,10 @@ pub fn execute_set_whitelist(
         return Err(ContractError::AlreadyStarted {});
     }
 
+    if config.whitelist.is_some() {
+        return Err(ContractError::WhitelistAlreadySet {});
```
JakeHartnell, line 195: So the intention here is to only let the whitelist be set once? No way to update it once it's been set?

JakeHartnell (approved):
LGTM

jhernandezb (approved):

PR #179, opened 2022-03-21
ICS 721 PR for view

humanalgorithm:


contracts/ics721/src/contract.rs
```diff
@@ -0,0 +1,220 @@
...
+        return Err(ContractError::NoSuchChannel { id: msg.channel });
+    };
+
+    // delta from user is in seconds
+    let timeout_delta = match msg.timeout {
+        Some(t) => t,
+        None => CONFIG.load(deps.storage)?.default_timeout,
+    };
+    // timeout is in nanoseconds
+    let timeout = env.block.time.plus_seconds(timeout_delta);
```
orkunkl, line 79:
```suggestion
    let timeout = env.block.time.plus_nanos(timeout_delta);
```

contracts/ics721/src/ibc.rs
```diff
@@ -0,0 +1,574 @@
...
+// TODO: copy pasta, make parent public
+// create a serialized error message
+fn ack_fail(err: String) -> Binary {
+    let res = Ics20Ack::Error(err);
+    to_binary(&res).unwrap()
+}
+
+const SEND_NFT_ID: u64 = 1338;
+
+#[cfg_attr(not(feature = "library"), entry_point)]
```
orkunkl, line 88: reply logic confused me for some time like, *is there an reply method for ibc?* 
Better to move this into contract.rs.

PR #187, opened 2022-03-24
Withdraw mint fees explicitly

### LINKED ISSUE #130: Withdraw function for mint fees
Minters should have to pay gas for funds to be distributed to creators. It's also good to have bit of a buffer period for funds to get distributed to creators, and it's good to make withdrawal explicit.

shanev:
Fixes #130 

Sends all mint fees to the contract, which can be withdrawn by the creator at a later date.

### DIFF
--- a/contracts/minter/src/contract.rs
+++ b/contracts/minter/src/contract.rs
@@ -433,16 +433,6 @@ fn _execute_mint(
     let new_mint_count = mint_count(deps.as_ref(), &info)? + 1;
     MINTER_ADDRS.save(deps.storage, info.clone().sender, &new_mint_count)?;
 
-    let mut seller_amount = Uint128::zero();
-    // Does have a fee
-    if !admin_no_fee {
-        seller_amount = mint_price.amount - network_fee;
-        msgs.append(&mut vec![CosmosMsg::Bank(BankMsg::Send {
-            to_address: env.contract.address.to_string(),
-            amount: vec![coin(seller_amount.u128(), config.unit_price.denom)],
-        })]);
-    };
-
     Ok(Response::default()
         .add_attribute("action", action)
         .add_attribute("sender", info.sender)
@@ -450,7 +440,6 @@ fn _execute_mint(
         .add_attribute("token_id", mintable_token_id.to_string())
         .add_attribute("network_fee", network_fee)
         .add_attribute("mint_price", mint_price.amount)
-        .add_attribute("seller_amount", seller_amount)
         .add_messages(msgs))
 }
 
--- a/contracts/minter/src/contract.rs
+++ b/contracts/minter/src/contract.rs
@@ -191,6 +191,9 @@ pub fn execute_withdraw(
     let balance = deps
         .querier
         .query_balance(env.contract.address, NATIVE_DENOM)?;
+    if balance.amount.is_zero() {
+        return Err(ContractError::ZeroBalance {});
+    }
 
     // send contract balance to creator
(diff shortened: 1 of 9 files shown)


contracts/minter/src/contract.rs
```diff
@@ -171,9 +171,38 @@ pub fn execute(
...
+        return Err(ContractError::Unauthorized(
+            "Sender is not an admin".to_owned(),
+        ));
+    };
+
+    // query balance from the contract
+    let balance = deps
+        .querier
+        .query_balance(env.contract.address, NATIVE_DENOM)?;
+
```
jhernandezb, line 194: should we check is greater than 0?

jhernandezb (changes requested):
LGTM besides one minor comment and needs to make the e2e test happy

contracts/minter/src/contract.rs
```diff
@@ -421,7 +450,7 @@ fn _execute_mint(
     if !admin_no_fee {
         seller_amount = mint_price.amount - network_fee;
         msgs.append(&mut vec![CosmosMsg::Bank(BankMsg::Send {
-            to_address: config.admin.to_string(),
+            to_address: env.contract.address.to_string(),
             amount: vec![coin(seller_amount.u128(), config.unit_price.denom)],
```
jhernandezb, line 454: do we really need this? funds are already in the contract

jhernandezb (changes requested):

contracts/minter/src/contract_tests.rs
```diff
@@ -1435,3 +1434,47 @@ fn unhappy_path() {
...
+    );
+    assert!(res.is_ok());
+
+    // Minter contract should have a balance
+    let minter_balance = router
+        .wrap()
+        .query_all_balances(minter_addr.clone())
+        .unwrap();
+    assert_eq!(minter_balance[0].amount.u128(), UNIT_PRICE - MINT_FEE);
+
```
JakeHartnell, line 1464: Maybe add a quick check that someone who is not the creator can't withdraw?

JakeHartnell (approved):
Like the direction!

One small comment about checking that non-creators can't withdraw contract balance in test.

contracts/minter/src/contract.rs
JakeHartnell: Yes.

contracts/minter/src/contract.rs
```diff
@@ -416,24 +448,13 @@ fn _execute_mint(
...
-    };
-
     Ok(Response::default()
         .add_attribute("action", action)
         .add_attribute("sender", info.sender)
         .add_attribute("recipient", recipient_addr)
         .add_attribute("token_id", mintable_token_id.to_string())
         .add_attribute("network_fee", network_fee)
         .add_attribute("mint_price", mint_price.amount)
-        .add_attribute("seller_amount", seller_amount)
```
jhernandezb, line 436: imo we shouldn't remove this attribute since will be used for data analysis, I guess its redundant and can be recomputed, but I think we should keep it so its consistent with the previous  version.

yubrew: Yes I agree

yubrew (approved):
just need to pass tests 👍

contracts/minter/src/contract.rs
shanev: Less work for the contract, the better imho (for a day when gas isn't zero). It can easily be computed by an indexer (mint_price - network_fee).

jhernandezb (approved):

PR #197, opened 2022-03-27
update base_token_uri and freeze

### LINKED ISSUE #180: Update and Lock baseTokenUri

yubrew:
fix #180 

1. Creator starts with a base token uri pointing to blank metadata pointing to the same initial teaser image. 
2. Minting happens.
3. Creator reveals later by updating base token uri pointing to final metadata and images.
4. Creator later freezes metadata so it can not be further updated.

- swapped `from` error messages around so `cw721_base::ContractError` is transformed into `ContractError`
- message type can be tricky. Needed to extend and then implement all the cw721_base execute msgs

### DIFF
--- a/contracts/sg721/src/contract.rs
+++ b/contracts/sg721/src/contract.rs
@@ -9,7 +9,6 @@ use sg_std::StargazeMsgWrapper;
 use crate::ContractError;
 use cw721::ContractInfoResponse;
 use cw721_base::ContractError as BaseError;
-use cw721_base::TokenInfo as Cw721TokenInfo;
 use url::Url;
 
 use crate::msg::{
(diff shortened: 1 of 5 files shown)


contracts/sg721/src/contract.rs
```diff
@@ -96,8 +99,74 @@ pub fn execute(
...
+        return Err(ContractError::Frozen {});
+    }
+
+    // iterate through all tokens
+    let all_tokens_msg = QueryMsg::AllTokens {
+        start_after: None,
+        limit: None,
+    };
+    let tokens_response: TokensResponse =
+        from_binary(&query(deps.as_ref(), env, all_tokens_msg).unwrap()).unwrap();
```
yubrew, line 150: This query is paginated, not sure how to handle it? Also code smell.

It's better to do `sg721` level change to store `base_token_uri` and return individual token uri as `{base_token_uri}/{token_id}`. Maybe as a separate issue.

yubrew:
closing in favor of implementing rand

JakeHartnell:
> closing in favor of implementing rand

Talking with Stewart... I think we should revive this for generative NFTs. Should be a separate contract, maybe `sg721-generative`?

yubrew:
I have gotten a couple pings for this. I'm not sure what the best approach
is, if we launch it as a separate sg721-generative minter or maybe make an
overarching minter with different "mint strategy" modules?

shanev:
> I have gotten a couple pings for this. I'm not sure what the best approach is, if we launch it as a separate sg721-generative minter or maybe make an overarching minter with different "mint strategy" modules?
> […](#)

I like this approach a lot. The first step is coming up with a common interface for strategies. Then it can be expressed as a trait which all strategies implement.

contracts/sg721/src/contract.rs
shanev: I think one way to do this is to make the `baseUri` optional. Then if it exists, update it only and treat the collection as if it's a generative collection. If `baseUri` doesn't exist, then use the individual `tokenUri`s.

shanev:
I think this should be a separate minter. Then creators can select what kind of minter to use.

yubrew:
I have the same sentiment. We need to keep all the shared group dependencies straight though.

shanev:
@yubrew Mind if I close this but keep the branch alive?

PR #204, opened 2022-04-02
Split Fair Burn into two functions

shanev:
Sometimes you want to call Fair Burn when you don't have access to `MessageInfo` (like in Marketplace contract). This adds a function for that.


jhernandezb (approved):
LGTM but I'd like to know the reasoning for this split.

PR #216, opened 2022-04-09
[WIP] SG2: Parameter Governance

shanev:


contracts/paramstore/src/contract.rs
```diff
@@ -0,0 +1,74 @@
...
+    match msg {
+        SudoMsg::UpdateParamu32 {
+            contract_name,
+            key,
+            value,
+        } => {
+            let param = U32.update(
+                deps.storage,
+                (contract_name.clone(), key.clone()),
+                |old| -> StdResult<_> { Ok(old.unwrap_or_default() + value) },
```
jhernandezb, line 47: is this doing an increment? I think it should only set the value right?

shanev: This should set the value if it doesn't exist or update it if it does. It currently doesn't do the latter. Have to write tests :)

PR #229, opened 2022-04-17
[WIP] Timelocked staking contract

shanev:


yubrew (commented):
this simplification in `instance per delegation` is really nice. avoids unnecessary complexity with map / end date edge cases.

shanev:
> this simplification in `instance per delegation` is really nice. avoids unnecessary complexity with map / end date edge cases.

It also gives a contract address per delegation, so delegation accounting can be entirely take care of by the chain's staking module and doesn't have to be duplicated in contracts.

PR #233, opened 2022-04-17
Add developer incentives to Fair Burn

shanev:
Fair Burn now takes an optional developer address where 10% of fees are distributed.

### DIFF
--- a/packages/sg1/src/lib.rs
+++ b/packages/sg1/src/lib.rs
@@ -1,37 +1,60 @@
-use cosmwasm_std::{coins, BankMsg, CosmosMsg, Decimal, MessageInfo, Uint128};
+use cosmwasm_std::{coins, Addr, BankMsg, CosmosMsg, Decimal, MessageInfo, Uint128};
 use cw_utils::{must_pay, PaymentError};
 use sg_std::{create_fund_community_pool_msg, StargazeMsgWrapper, NATIVE_DENOM};
 use thiserror::Error;
 
 // governance parameters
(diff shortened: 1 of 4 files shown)


packages/sg1/src/lib.rs
```diff
@@ -1,37 +1,60 @@
...
+    let (burn_percent, dev_fee) = match developer {
+        Some(dev) => {
+            let dev_fee = (Uint128::from(fee) * Decimal::percent(DEV_INCENTIVE_PERCENT)).u128();
+            let msg = BankMsg::Send {
+                to_address: dev.to_string(),
+                amount: coins(dev_fee, NATIVE_DENOM),
+            };
+            msgs.push(SubMsg::Bank(msg));
+            (
+                Decimal::percent(FEE_BURN_PERCENT - DEV_INCENTIVE_PERCENT),
```
yubrew, line 39: what if `DEV_INCENTIVE_PERCENT > FEE_BURN_PERCENT`

yubrew (commented):
adding these incentives looks very simple

packages/sg1/src/lib.rs
shanev: This should be enforced in the governance layer (when its implemented). Right now its fine since the parameters are hardcoded.

shanev:
> adding these incentives looks very simple

Hopefully we can find a way to add these to strategies so contracts won't have to be forked to add incentives.

yubrew (approved):

PR #240, opened 2022-04-24
Handle finalize sale hook for NFT buy airdrop claim

### LINKED ISSUE #239: Claim should respond to buy NFT hook

shanev:
Fixes #239 

Includes our first controller (for hooks). Controllers are re-usable contract components that encapsulate common functionality. The usually wrap around a storage element (like `Item`) and add some logic.

### DIFF
--- a/contracts/claim/src/contract.rs
+++ b/contracts/claim/src/contract.rs
@@ -114,6 +114,3 @@ pub fn execute_claim_buy_nft(
 pub fn query(_deps: Deps, _env: Env, _msg: QueryMsg) -> StdResult<Binary> {
     Ok((to_binary(&"queries not implemented".to_string())).unwrap())
 }
-
-// TODO: add tests for hook after publishing Marketplace code
-// https://github.com/public-awesome/stargaze-contracts/issues/241
--- a/contracts/claim/src/contract.rs
+++ b/contracts/claim/src/contract.rs
@@ -8,6 +8,7 @@ use sg_std::{create_claim_for_msg, ClaimAction, StargazeMsgWrapper};
 
 use crate::error::ContractError;
 use crate::msg::{ExecuteMsg, InstantiateMsg, SaleFinalizedHookMsg};
+use crate::state::{Config, MarketplaceContract, CONFIG};
 pub type Response = cosmwasm_std::Response<StargazeMsgWrapper>;
 
 // version info for migration info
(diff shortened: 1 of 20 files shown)


contracts/claim/src/multitest.rs
```diff
@@ -0,0 +1,255 @@
...
+            token_id: TOKEN_ID,
+            expires: router.block_info().time.plus_seconds(MIN_EXPIRY + 1),
+        };
+        let _res = router.execute_contract(
+            bidder,
+            marketplace_addr,
+            &set_bid_msg,
+            &coins(100, NATIVE_DENOM),
+        );
+        // TODO: this fails, maybe multitest doesn't support hooks yet?
```
JakeHartnell, line 241: Hmmmmm, feel this should work...

shanev: Fixed.

contracts/claim/src/contract.rs
```diff
@@ -19,9 +22,20 @@ pub fn instantiate(
     deps: DepsMut,
     _env: Env,
     _info: MessageInfo,
-    _msg: InstantiateMsg,
+    msg: InstantiateMsg,
 ) -> Result<Response, ContractError> {
+    let marketplace =
+        MarketplaceContract(deps.api.addr_validate(&msg.marketplace_addr).map_err(|_| {
```
jhernandezb, line 28: why does this need to depend on sg_marketplace and not just be an address ? imo claim contract shouldn't care as long as it has an "admin"

e2e/claim_test.go
```diff
@@ -82,15 +84,57 @@ func TestClaim(t *testing.T) {
...
+		CodeID: 4,
+		Label:  "Marketplace",
+		Msg:    instantiateMsgRaw,
+	})
+	require.NoError(t, err)
+	require.NotNil(t, instantiateRes)
+	require.NotEmpty(t, instantiateRes.Address)
+	marketplaceAddress := instantiateRes.Address
+	require.NotEmpty(t, marketplaceAddress)
+
```
jhernandezb, line 130: I guess the e2e testing for buying an nft would be pending? we can track it on a different issue

jhernandezb (changes requested):

contracts/claim/src/contract.rs
yubrew: Adding to this, would be nice to have a generalized contract addr checker

shanev: @jhernandezb It doesn't need to be but I wanted to start using contract helpers within contracts instead of raw messages for convenience. This contract depends on sg_marketplace for the hook anyway. It is a bit redundant to use the helper here so can remove it if you feel strongly about it.

e2e/claim_test.go
shanev: @jhernandezb https://github.com/public-awesome/marketplace/issues/91

yubrew (approved):
👍

contracts/claim/src/contract.rs
jhernandezb: sounds good to me

jhernandezb (approved):

PR #261, opened 2022-05-06
Fix distribution mechanism on royalty-group contract

tasiov:
Adjusting the `execute_distribute_funds` function so that it sends a proportional amount to each member in the group. The proportion is determined by dividing a members weight by the total weight of all members.

Added test coverage as well.


tasiov:
Hmm I think I misunderstood the intention of the contract here.

shanev (approved):
🎉

shanev:
Thanks! The next step is the ability to send funds directly to the contract, and allow members to distribute it to themselves (or an admin to distribute to all): https://github.com/public-awesome/stargaze-contracts/issues/265.

PR #271, opened 2022-05-24
Adding the ability to lock minting for users

findolor:
The `mint_to` and `mint_for` messages only allow for admin to execute the contract. But in `mint_sender`, anyone can call the contract and execute the minting.

For some of the projects with extra logic this can be a problem if they don't want users to mint tokens on their own.

Added a lock variable for making the `mint_sender` dynamically change the minting permission. The contract still allows users to mint freely without explicitly setting the `lock_minter` field so unless users update this field or instantiate it as true, it is the same as before - everyone can mint freely.


findolor:
Happy to contribute! 

For the lock, it's not really about the minting time but a different approach to logic.

Let's say that I have a contract that has some game logic. I want to mint some NFTs to people on some condition and not let them mint on their own. This lock could be used to achieve that functionality.

shanev:
While an interesting feature, there hasn't been any requests for this functionality so far. Don't want to add too many things until they're needed. I say we keep this branch alive for now and add it when/if its needed. That cool?

findolor:
@shanev Sounds good to me, thanks!

PR #289, opened 2022-05-29
add e2e readme

yubrew:


jhernandezb:
Seems a bit like this https://github.com/public-awesome/stargaze-contracts#running-e2e-tests

no harm on having its own README tho since will be easier to find.

jhernandezb (approved):

PR #290, opened 2022-05-31
[WIP] token index

yubrew:
saving the shuffled map doesn't make a difference since it'll be hashed the same way in the merkle tree so querying order will not change

requires using an index where value holds the token id.

to know which token key to remove after minting, value also needs to hold token key.


yubrew:
using wrong base, remaking with `working token index`

PR #291, opened 2022-05-31
working token index

yubrew:
saving the shuffled map doesn't make a difference since it'll be hashed the same way in the merkle tree so querying order will not change. requires using an index. data needs both `token_key` and `token_id`. to both query the presence of `token_id` and know which `token_key` to remove.

please double check token key and token id references are correct.

### DIFF
--- a/contracts/minter/src/contract.rs
+++ b/contracts/minter/src/contract.rs
@@ -6,7 +6,7 @@ use crate::msg::{
     MintableNumTokensResponse, QueryMsg, StartTimeResponse,
 };
 use crate::state::{
-    Config, CONFIG, MINTABLE_NUM_TOKENS, MINTABLE_TOKEN_IDS, MINTER_ADDRS, SG721_ADDRESS,
+    Config, CONFIG, MINTER_ADDRS, SG721_ADDRESS, tokens, TokenInfo, increment_tokens, decrement_tokens, token_count,
 };
 #[cfg(not(feature = "library"))]
 use cosmwasm_std::entry_point;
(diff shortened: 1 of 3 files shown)


contracts/minter/src/contract.rs
```diff
@@ -120,12 +121,19 @@ pub fn instantiate(
         start_time: msg.start_time,
     };
     CONFIG.save(deps.storage, &config)?;
-    MINTABLE_NUM_TOKENS.save(deps.storage, &msg.num_tokens)?;
 
     let token_list = random_token_list(&env, msg.num_tokens)?;
     // Save mintable token ids map
     for token_id in token_list {
-        MINTABLE_TOKEN_IDS.save(deps.storage, token_id, &true)?;
+        let count = increment_tokens(deps.storage)?;
```
jhernandezb, line 128: I made the same comment on Shane's PR, I think  can we avoid saving the counter every time and just do it once the loop ends and use the  index (index,token_id) to save it in TokenInfo? for 10k collections write operations will have some gas costs implications

contracts/minter/src/state.rs
```diff
@@ -18,6 +18,44 @@ pub struct Config {
...
+}
+
+impl<'a> IndexList<TokenInfo> for TokenIndices<'a> {
+    fn get_indexes(&'_ self) -> Box<dyn Iterator<Item = &'_ dyn Index<TokenInfo>> + '_> {
+        let v: Vec<&dyn Index<TokenInfo>> = vec![&self.token_ids];
+        Box::new(v.into_iter())
+    }
+}
+
+pub fn tokens<'a>() -> IndexedMap<'a, TokenKey, TokenInfo, TokenIndices<'a>> {
```
shanev, line 56:
```suggestion
pub fn mintable_tokens<'a>() -> IndexedMap<'a, TokenKey, TokenInfo, TokenIndices<'a>> {
```

contracts/minter/src/contract.rs
```diff
@@ -124,8 +125,17 @@ pub fn instantiate(
 
     let token_list = random_token_list(&env, msg.num_tokens)?;
     // Save mintable token ids map
-    for token_id in token_list {
-        MINTABLE_TOKEN_IDS.save(deps.storage, token_id, &true)?;
+    for (i, token_id) in token_list.iter().enumerate() {
+        // i = [0..num_tokens], add 1 to get the expected token key
+        let token_key = i as u32 + 1;
```
shanev, line 130: Do you need to cast as `u32` here?

shanev (changes requested):
Just some naming suggestions.

contracts/minter/src/contract.rs
yubrew: yes or else i is usize, wrong type

yubrew:
takes up too much gas. couldn't even fit on a block

PR #294, opened 2022-06-01
pub shuffle

### LINKED ISSUE #292: public shuffle
allow public to shuffle mint order

yubrew:
fix #292

please don't merge until removing the mintable tokens query

### DIFF
--- a/contracts/minter/src/contract_tests.rs
+++ b/contracts/minter/src/contract_tests.rs
@@ -1549,7 +1549,7 @@ fn shuffle() {
     // try shuffle
     let shuffle_msg = ExecuteMsg::Shuffle {};
     let funds = coins(SHUFFLE_FEE, NATIVE_DENOM);
-    let res = router.execute_contract(creator.clone(), minter_addr.clone(), &shuffle_msg, &funds);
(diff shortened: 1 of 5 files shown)


contracts/minter/src/contract_tests.rs
```diff
@@ -1462,6 +1462,97 @@ fn unhappy_path() {
...
+    // setup contracts
+    let num_tokens = 20;
+    let (minter_addr, _config) = setup_minter_contract(&mut router, &creator, num_tokens);
+    // query mintable order for mints
+    let query_mintable_tokens_msg = QueryMsg::MintableTokens {};
+    let res: MintableTokensResponse = router
+        .wrap()
+        .query_wasm_smart(minter_addr.clone(), &query_mintable_tokens_msg)
+        .unwrap();
+    println!("{:?}", res);
```
penso, line 1481: Isn't it nicer to use something like `log::debug!()` instead of `println!`? Just wondering.

jhernandezb: I have used ```dbg!()``` macro

penso: `dbg!()` sounds like a good option too

yubrew:
closing, part of this code was used in a different pull

PR #299, opened 2022-06-06
[WIP] shuffle array

yubrew:
hierarchy of pulls:
pull/208 "yubrew/132-implement-rand"
-pull/221 "jhernandezb/132-implement-rand-list"
-- pull/291 "yubrew/working-token-index"
--- pull/294 "yubrew/pub-shuffle"
---- pull/299 "yubrew/shuffle-array"

if pull 299 works then we're gucci

random minter tests:

- try minter instantiate look at tx for gas
- try random minting and look at tx's for gas
- try shuffle and look at tx's for gas
- try random minting and look at tx's for gas


yubrew:
array is cheaper instantiation, but because of item load -> save, incurs too much gas for mint

PR #304, opened 2022-06-16
remove unneeded binary export 

pyramation:
latest cosmwasm-typescript-gen upgrade fixes an issue that required us to manually export the Binary type. Not necessary anymore, so this configuration is more minimal.


shanev (approved):

shanev:
Thanks!

PR #314, opened 2022-06-23
randomized token map { position: id }

yubrew:
**DO NOT MERGE YET** need to remove the mintable tokens query before merging

saving the shuffled map doesn't make a difference since it'll be hashed the same way in the merkle tree so querying order will not change. requires using an index. data needs both `token_key` and `token_id`. to both query the presence of `token_id` and know which `token_key` to remove.

this is the v0.1 PR resulting from a series of tests
- switch tokens to multi map
- switch tokens to array
- benchmark against existing token instantiation
- move validation to the front of function blocks

we settled on this version because it adds cost efficient randomization and a few additional enhancements.

still need to test a few things on testnet:
- try minter instantiate look at tx for gas
- try random minting and look at tx's for gas
- try shuffle and look at tx's for gas
- try random minting and look at tx's for gas
- try mint_for multiple times and inspect token ids
- try mint_to multiple times and inspect token ids

```
TESTNET CODE_ID: 91
minter: stars15gdrt02y3x6juldarcu2rt6y6l8gklaneh446uj2yrcsa7myamtqj5zlgf
sg721:  stars175evjsxpxf5347an078zmk3gzjy7df6kjml0mfwxvzpuu5jps8gsx3caxe
batch mint 20 tokens takes 37M gas https://stargaze-testnet-explorer.pages.dev/stargaze/tx/A72BAD4520B2E23DEE2844B4A735FC33931588A86C414EF200A978DEC05A6F8F
```

### DIFF
--- a/contracts/minter/src/contract.rs
+++ b/contracts/minter/src/contract.rs
@@ -125,12 +125,12 @@ pub fn instantiate(
     CONFIG.save(deps.storage, &config)?;
     MINTABLE_NUM_TOKENS.save(deps.storage, &msg.num_tokens)?;
 
-    let token_positions = random_token_list(&env, (1..=msg.num_tokens).collect::<Vec<u32>>())?;
+    let token_ids = random_token_list(&env, (1..=msg.num_tokens).collect::<Vec<u32>>())?;
     // Save mintable token ids map
-    let mut token_id: u32 = 1;
-    for token_position in token_positions {
+    let mut token_position: u32 = 1;
+    for token_id in token_ids {
         MINTABLE_TOKEN_POSITIONS.save(deps.storage, token_position, &token_id)?;
-        token_id += 1;
+        token_position += 1;
     }
 
     // Submessage to instantiate sg721 contract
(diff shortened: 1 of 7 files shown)


contracts/minter/src/contract.rs
```diff
@@ -455,15 +510,14 @@ fn _execute_mint(
...
         .add_attribute("network_fee", network_fee)
         .add_attribute("mint_price", mint_price.amount)
         .add_messages(msgs))
 }
 
-fn random_token_list(env: &Env, num_tokens: u32) -> Result<Vec<u32>, ContractError> {
-    let mut tokens: Vec<u32> = (1..=num_tokens).collect::<Vec<u32>>();
-    let sha256 = Sha256::digest(format!("{}{}", env.block.height, num_tokens).into_bytes());
+fn random_token_list(env: &Env, mut tokens: Vec<u32>) -> Result<Vec<u32>, ContractError> {
+    let sha256 = Sha256::digest(format!("{}{}", env.block.height, tokens.len()).into_bytes());
```
shanev, line 520: Should we add more to this like sender?

shanev (approved):

contracts/minter/src/contract.rs
yubrew: sure

PR #320, opened 2022-06-27
Dynamic collection creation fee

### LINKED ISSUE #318: Dynamic collection creation cost

shanev:
Fixes #318

### DIFF
--- a/contracts/minter/src/contract.rs
+++ b/contracts/minter/src/contract.rs
@@ -135,6 +135,7 @@ pub fn instantiate(
                 symbol: msg.sg721_instantiate_msg.symbol,
                 minter: env.contract.address.to_string(),
                 collection_info: msg.sg721_instantiate_msg.collection_info,
+                num_tokens: None,
             })?,
             funds: info.funds,
             admin: Some(info.sender.to_string()),
--- a/contracts/sg721/src/contract.rs
+++ b/contracts/sg721/src/contract.rs
@@ -3,6 +3,8 @@ use cosmwasm_std::entry_point;
 use cosmwasm_std::{to_binary, Binary, Deps, DepsMut, Empty, Env, MessageInfo, StdResult};
 use cw2::set_contract_version;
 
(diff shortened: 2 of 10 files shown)


contracts/sg721/src/contract.rs
```diff
@@ -3,6 +3,8 @@ use cosmwasm_std::entry_point;
 use cosmwasm_std::{to_binary, Binary, Deps, DepsMut, Empty, Env, MessageInfo, StdResult};
 use cw2::set_contract_version;
 
+use rust_decimal::prelude::ToPrimitive;
+use rust_decimal::Decimal;
```
JakeHartnell, line 7: Curious why we're using `rust_decimal` here rather than `Decimal` from `cosmwasm_std`?

shanev: Because CW `Decimal` is just a wrapper around `String` and doesn't provide `ceil()` which this logic uses.

JakeHartnell: Makes sense, that is elegant.

Last thought, otherwise LGTM... do we want the creation fee to be "per 1000"? If `CREATION_FEE_PER_1000` is 5000 STARS, why not just charge 5 STARS per token?

Though now that I think about it, `CREATION_FEE_PER_1000` would certainly bring in more fee revenue. :thinking:

contracts/minter/src/contract.rs
```diff
@@ -135,6 +135,7 @@ pub fn instantiate(
                 symbol: msg.sg721_instantiate_msg.symbol,
                 minter: env.contract.address.to_string(),
                 collection_info: msg.sg721_instantiate_msg.collection_info,
+                num_tokens: None,
```
yubrew, line 138: i don't understand why you're adding this in here. is it to make dynamic collection fees optional?

shanev:
Will be superseded by Launchpad v2 minter factory governance.

PR #321, opened 2022-06-28
Limit per address limit to 1%

### LINKED ISSUE #296: Max mint limit based on collection size

shanev:
Fixes #296


contracts/minter/src/contract.rs
```diff
@@ -68,6 +68,17 @@ pub fn instantiate(
         });
     }
 
+    // Limit per address limit to 1% of num tokens
+    if (msg.num_tokens > 100 && msg.num_tokens < 5000)
```
yubrew, line 72: i don't like this conditional because `msg.num_tokens <= 100` => `per_address_limit` could be `50`.
maybe use ceil(dec(1/100))? i'm not sure

contracts/vending-minter/src/contract.rs
```diff
@@ -65,6 +65,17 @@ pub fn instantiate(
         .querier
         .query_wasm_smart(factory.clone(), &Sg2QueryMsg::Params {})?;
 
+    // Limit per address limit to 1% of num tokens
+    if (msg.init_msg.num_tokens > 100 && msg.init_msg.num_tokens < 5000)
+        && msg.init_msg.per_address_limit > (msg.init_msg.num_tokens / 100)
```
shanev, line 70: Having trouble understanding this logic and hardcoded values. A comment could help.

contracts/vending-minter/src/contract.rs
```diff
@@ -621,6 +632,18 @@ pub fn execute_update_per_address_limit(
             got: per_address_limit,
         });
     }
+
+    // Limit per address limit to 1% of num tokens
+    if (config.extension.num_tokens > 100 && config.extension.num_tokens < 5000)
```
shanev, line 637: Duplicate code? Could possibly break out into a self-describing function.

PR #333, opened 2022-07-03
update mint price

### LINKED ISSUE #300: Update mint price before mint starts

yubrew:
fix #300


shanev (approved):
LGTM, just have to be updated.

contracts/vending-minter/src/contract.rs
```diff
@@ -590,6 +592,45 @@ fn random_mintable_token_mapping(
...
+    }
+
+    // TODO add test
+    // If price is greater than initial price, return error
+    if price > config.extension.initial_price.amount.u128() {
+        return Err(ContractError::UpdatedMintPriceTooHigh {
+            allowed: config.extension.initial_price.amount.u128(),
+            updated: price,
+        });
+    }
```
jhernandezb, line 624: This won't let update price above the initial price even if the public start time hasn't started.

jhernandezb (changes requested):

contracts/vending-minter/src/contract.rs
yubrew: is there a point to `initial_price` if you can set it above or below before the sale and it's not used after sale?

contracts/vending-minter/src/contract.rs
```diff
@@ -590,6 +591,36 @@ fn random_mintable_token_mapping(
...
+    price: u128,
+) -> Result<Response, ContractError> {
+    nonpayable(&info)?;
+    let mut config = CONFIG.load(deps.storage)?;
+    if info.sender != config.extension.admin {
+        return Err(ContractError::Unauthorized(
+            "Sender is not an admin".to_owned(),
+        ));
+    }
+    // TODO add test
```
shanev, line 607:
```suggestion
```

shanev (approved):

jhernandezb (approved):

PR #350, opened 2022-08-19
Use sg721-base rather than cw721-base

JakeHartnell:
@yubrew

Here's an example of how to use `sg721-base` with `sg721-nt`. It's a little confusing since we're extending an extension... may inform some refactoring.

### DIFF
--- a/contracts/sg721-nt/src/lib.rs
+++ b/contracts/sg721-nt/src/lib.rs
@@ -14,8 +14,8 @@ pub mod entry {
         contract::_instantiate,
         msg::{ExecuteMsg, QueryMsg},
     };
-    use cosmwasm_std::{to_binary, Binary, Deps, DepsMut, Env, MessageInfo, StdResult};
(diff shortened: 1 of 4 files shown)


contracts/sg721-nt/src/lib.rs
```diff
@@ -3,11 +3,10 @@ use cosmwasm_std::entry_point;
 
 pub mod contract;
 pub mod msg;
-use cw721_base::Extension;
 use sg721::InstantiateMsg;
-use sg_std::StargazeMsgWrapper;
+pub use sg721_base::Extension;
 
-pub type Cw721Base<'a> = cw721_base::Cw721Contract<'a, Extension, StargazeMsgWrapper>;
+pub type Sg721Base<'a> = sg721_base::Cw721Base<'a>;
```
JakeHartnell, line 9: This is the tricky bit... may want to rename this in `sg721-base`. This type already includes the extension and msg wrapper which is nice.

yubrew (approved):
👍

PR #351, opened 2022-08-20
Factory + minter refactoring

shanev:
Fixes most of the comments from @yubrew code review.

* Move minter status from factory -> minter
* Remove factory controller in favor of common code in base factory
* Remove reply that's not needed after the minter status change


contracts/base-minter/src/contract.rs
```diff
@@ -42,6 +41,9 @@ pub fn instantiate(
 
     let factory = info.sender.clone();
 
+    // set default status so it can be queried without failing
+    STATUS.save(deps.storage, &Status::default())?;
```
yubrew, line 45: is default of bool `false`?

shanev: Yeah the default is `false`. But since the `STATUS` storage is in the contract, the item has to be saved at least once on init to exist and not fail `.load()` queries.

PR #357, opened 2022-08-26
start trading time

### LINKED ISSUE #352: Add start trading param

yubrew:
fix #352


contracts/base-minter/src/contract.rs
```diff
@@ -56,7 +58,12 @@ pub fn instantiate(
         // assume the mint price is the minimum mint price
         // 100% is fair burned
         mint_price: res.params.min_mint_price,
-        extension: Empty {},
+        start_trading_time: None,
+        extension: ConfigExtension {
```
shanev, line 62: Is the creator always the contract admin? i.e: Can't this creator have a different account than the person responsible for instantiating and updating the collection?

Been wanting to refactor the admin stuff a bit. Basically wanted to keep the base minter from having to use any extensions. Is the `admin` from `ContractInfoResponse` the right one here? (see https://github.com/CosmWasm/cosmwasm/blob/3740c0739ba80fdbabecc5c47f94b4be01f40294/packages/std/src/query/wasm.rs#L36). If not, maybe a `creator` can be part of the default setup.

yubrew: Creator may leave the project down the line. taking `admin` from `ContractInfoResponse` seems better maintained. Using this could leave an empty extension and allow creators flexibility in setting the `start_trading_time`.

PR #365, opened 2022-08-31
Splits with tests

shanev:
Adds a new contract sg-splits, which simply distributes funds from the contract when a distribute function is called. It is initialized with a cw4-group, which manages the membership and weight for each member.

Vending minter is updated to work with paying a contract. It also supports paying to a DAO. An optional `payment_address` was added to the params to make alternate payment methods more flexible.


contracts/vending-minter/src/contract.rs
```diff
@@ -523,6 +529,48 @@ fn _execute_mint(
...
+// If the payment address is a splits contract, perform the split. Else, pay single seller.
+fn pay_seller(
+    deps: Deps,
+    amount: Uint128,
+    mint_price: Coin,
+    payment_address: Option<Addr>,
+    creator: Addr,
+    res: &mut Response,
+) -> Result<(), ContractError> {
+    if let Some(addr) = payment_address {
```
shanev, line 541: Handle this logic correctly.

PR #370, opened 2022-09-02
fix dependencies to pass tests

yubrew:
using rel path instead of versions... can use abs versions before compiling wasms and gov vote


shanev (approved):
This is fine. Will revert back before publishing the main update.

PR #375, opened 2022-09-07
start trading time

### LINKED ISSUE #352: Add start trading param

yubrew:
fix #352 and more


shanev (approved):

contracts/sg721-base/src/contract.rs
```diff
@@ -214,6 +221,82 @@ where
...
+
+        if let Some(ref external_link) = new_collection.external_link {
+            Url::parse(external_link)?;
+        }
+
+        if let Some(start_trading_time) = new_collection.start_trading_time {
+            if env.block.time > start_trading_time {
+                return Err(ContractError::InvalidStartTradingTime {});
+            }
+        }
```
jhernandezb, line 258: start trading time should only be be updated  through the minter because it needs to check public mint start time + offset, also it should only allow to decrease it after public mint starts and this will allow to just keep extending it until is frozen.

contracts/sg721-base/src/contract.rs
```diff
@@ -67,13 +67,16 @@ where
             description: msg.collection_info.description,
             image: msg.collection_info.image,
             external_link: msg.collection_info.external_link,
+            start_trading_time: None,
```
jhernandezb, line 70: can't be set during instantiation? doesn't seem to be taking the information from the `InstantiateMsg`

contracts/sg721-base/src/contract.rs
```diff
@@ -214,6 +221,82 @@ where
...
+        if frozen_collection_info {
+            return Err(ContractError::CollectionInfoFrozen {});
+        }
+
+        let collection = self.query_collection_info(deps.as_ref())?;
+        if collection.creator != info.sender {
+            return Err(ContractError::Unauthorized {});
+        }
+
+        deps.api.addr_validate(&new_collection.creator)?;
```
jhernandezb, line 242: do we really want to change the creator? nothing stops this from impersonating another creator.

jhernandezb (changes requested):

contracts/base-minter/src/contract.rs
```diff
@@ -161,6 +174,51 @@ pub fn execute_mint_sender(
...
+    // add custom rules here
+    if let Some(start_time) = start_time {
+        // If current time already passed the new start_time return error
+        if env.block.time > start_time {
+            return Err(ContractError::InvalidTradingStartTime(
+                env.block.time,
+                start_time,
+            ));
+        }
+    }
```
jhernandezb, line 205: missing check 
```
start_time >  (env.block.time + max_period)
  return ERR
```

contracts/vending-minter/src/contract.rs
```diff
@@ -644,6 +662,48 @@ pub fn execute_update_start_time(
...
+
+    // add custom rules here
+    if let Some(start_time) = start_time {
+        // If current time already passed the new start_time return error
+        if env.block.time > start_time {
+            return Err(ContractError::InvalidTradingStartTime(
+                env.block.time,
+                start_time,
+            ));
+        }
```
jhernandezb, line 689: same validation needed here.

jhernandezb (changes requested):

jhernandezb (approved):

PR #379, opened 2022-09-10
Upgrade ts-codgen and update Typescript types 

shanev:
Build types with:

```
cd ts
yarn codegen
```

## NOTE: I'm not sure if Galaxy uses the old types. If so, it may need to be updated to use this.


yubrew (approved):
👍 UTAck

pyramation:
this is pretty cool! I do think we should consider making it another repository instead of baking everything into this one?

or if you want to keep it here, let me help dial some of the configs ;)

shanev:
@pyramation Yeah I think it should be a separate repo. So should I copy this stuff to a new repo or fork the one at cosmology-tech?

pyramation:
> @pyramation Yeah I think it should be a separate repo. So should I copy this stuff to a new repo or fork the one at cosmology-tech?

I can send the cosmology-tech one over if you want?

shanev:
Merging this in for now but will break it out into a separate repo son.

PR #388, opened 2022-09-16
Create group within splits

shanev:
This creates the group within the splits contract instead of passing it in. This makes the frontend flow a lot easier as you don't have to 1) create a group, 2) search for the group, 3) create splits with the group address.

To discuss: who should be the admin of the group contract, if any? In this PR it's set to the sender.


shanev:
Replaced by #492

PR #395, opened 2022-09-19
check trading start time

### LINKED ISSUE #390: check trading start time doesn't go above offset
trading_start_time > (public start + offset) then error

yubrew:
fix #390

### DIFF
--- a/contracts/base-minter/src/contract.rs
+++ b/contracts/base-minter/src/contract.rs
@@ -194,20 +194,10 @@ pub fn execute_update_trading_start_time(
     }
 
     // add custom rules here
-    let config = CONFIG.load(deps.storage)?;
-    let factory_params: ParamsResponse = deps
-        .querier
-        .query_wasm_smart(config.factory, &Sg2QueryMsg::Params {})?;
-    let default_start_time_with_offset = env
-        .block
-        .time
-        .plus_seconds(factory_params.params.default_trading_offset_secs);
-
     if let Some(start_time) = start_time {
-        // If current time already passed the new start_time return error
-        if default_start_time_with_offset > start_time {
+        if env.block.time > start_time {
             return Err(ContractError::InvalidTradingStartTime(
-                default_start_time_with_offset,
+                env.block.time,
                 start_time,
             ));
         }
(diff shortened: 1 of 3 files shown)


shanev (approved):

contracts/base-minter/src/contract.rs
```diff
@@ -194,11 +194,20 @@ pub fn execute_update_trading_start_time(
...
     if let Some(start_time) = start_time {
         // If current time already passed the new start_time return error
-        if env.block.time > start_time {
+        if default_start_time_with_offset > start_time {
             return Err(ContractError::InvalidTradingStartTime(
-                env.block.time,
+                default_start_time_with_offset,
                 start_time,
             ));
         }
```
jhernandezb, line 213: I don't think this is working as expected. The end goal here is to prevent setting the trading start time beyond than public mint time + offset, public mint start time can not be changed after it starts so this ensure it gets frozen. 

So this needs 2 checks. 

```
// 1 cannot set it to a time in the past
if env.block.time > start_time {
  return err
}

// 2-  cannot go beyond the offset 
if start_time > (minter.start_time + offset){
   return err
}
```

contracts/vending-minter/src/contract.rs
```diff
@@ -679,11 +679,19 @@ pub fn execute_update_trading_start_time(
...
-        // If current time already passed the new start_time return error
-        if env.block.time > start_time {
+        // If old start time + offset > new start_time, return error
+        if default_start_time_with_offset > start_time {
             return Err(ContractError::InvalidTradingStartTime(
-                env.block.time,
+                default_start_time_with_offset,
                 start_time,
             ));
         }
```
jhernandezb, line 697: Same logic as above.

jhernandezb (changes requested):

contracts/base-minter/src/contract.rs
yubrew: Base minter does not have `start time` because it's intended for the creator to self mint. That's what makes this one more confusing.

Maybe case 1 is sufficient?

contracts/vending-minter/src/contract.rs
yubrew: This makes sense to change.

contracts/base-minter/src/contract.rs
jhernandezb: Keep forgetting  about base minter is 1-1 lol, you are right its only needed for base minter then.

jhernandezb (approved):

PR #396, opened 2022-09-19
prune map

### LINKED ISSUE #391: Prune mint count after sold out
prune mint count map after is fully minted

yubrew:
fix #391


shanev (approved):

jhernandezb (approved):
Looks good to me, maybe an optional .take(N) would be nice. so we can batch deletes if the map is too big to do in a single tx.

PR #416, opened 2022-09-23
add explicit to collection info

### LINKED ISSUE #408: Add an explicit flag to collection metadata
So creators can mark their collections as explicit. These could be hidden in the UI but revealed in filtering.

yubrew:
fix #408


jhernandezb (approved):
Looks good!

PR #427, opened 2022-10-05
check whitelist on instantiation and on set

jhernandezb:


yubrew (approved):
i'd like to see 1 more additional test for `WhitelistAlreadyStarted`. the rest looks good

jhernandezb:
added test for whitelist already started

Issue #131, opened 2022-03-05
Merkle whitelist


shanev:
Minters should have to pay gas for funds to be distributed to creators. It's also good to have bit of a buffer period for funds to get distributed to creators, and it's good to make withdrawal explicit.

shanev:
Devil's advocate: It's actually a benefit of Stargaze to store whitelists on-chain. You can't do this on Ethereum because of storage and gas constraints. But Stargaze doesn't have those issues.

Issue #373, opened 2022-09-04
Use merkletree for lists


yubrew:
Such as whitelist

PR #429, opened 2022-10-09
Merkle Tree whitelist

0xArbi:
I saw the [Use merkletree for lists](https://github.com/public-awesome/launchpad/issues/373) issue had been opened recently, and thought it's 

1. a very cool feature for mass/institutional adoption of Stargaze NFTs, I'm thinking of Microsoft or another institution dropping a million NFTs to their existing users. Right now I believe whitelists are limited to 5_000 entries
2. a great opportunity to build up my Rust skills a little further.

Apologies if there was already someone looking into this or I'm way off in the direction you wanted to take. 

### Approach

- Created a new `Whitelist` contract called `MerkleWhitelist`
  - Added a small TS package for creating test merkle trees. Pretty much copied from [cw20-merkle-airdrop](https://github.com/CosmWasm/cw-tokens/tree/main/contracts/cw20-merkle-airdrop/src) 
- Updated the minter to now takes a `Whitelist` object that encodes the type (ie. `List` or `Merkle`)
- Updated the tests to accomodate the new type of `Whitelist`

### Questions

I'm still a Rust/CosmWasm noob so I have some questions,

- Given merkle trees can encode many items, what should the WL creation fee be? Still related to the number of entries or fixed?
- Right now I've decided to make a new WL contract, pushing complexity into the minter contract. Is the best approach to, 
  1. leave as is
  2. to create a new `MerkleMinter`
  3. make the existing `Whitelist` contract support merkle trees
- I'm super interested to see the gas impact of minting when the tree has say, 50_000 to 1_000_000 entries. How can I do gas metering locally?
- I can't quite fathom how upgrades to CosmWasm contracts work just yet. I understand there's probably 1 canonical deployment of the minter contract, so is that all that needs to be migrated? 
- I don't quite like the `num_tokens_per_address` field in relation to the merkle WL, it'd be nicer to encode the number of items per address in the tree itself, thoughts?


yubrew:
Thanks for contributing this! I'm going to chime in first with potentially wrong answers that hopefully someone downstream will correct. 

We haven't finished this feature yet but thinking to tie costs to a stars per byte storage fee. Then gut check that against usd equivalent to avoid spam on a gree gas cost chain.

I think there can be a whitelist v2 that accommodates merkletrees and maybe change the interface for a `member_in_whitelist(addr)` with `bool` response that handles both normal list of addrs and merkletree.

We can store the new whitelist v2 contract, retire the old contract. migrations are not bad if there are no state changes.

There are probably better ways to do this. I would store the contract and instantiate on testnet (permissionless) with a minter. Then you can do side by side comparisons on the gas usage for the different txs.

We did it the map way because it was easy to reason and we did not explore merkletrees as a data structure. If you store mint count in the tree that sounds like a big storage savings.

Also we are thinking to add `purge` after contracts expire (minter, whitelist) to reduce blockchain state bloat.

We will look through the code hopefully later next week.

yubrew:
fix #131 #373

shanev:
Would love someone to research this further and get into a production-ready state.

PR #437, opened 2022-10-18
Minor tidy-up

larry0x:
* Move `go.mod` and `go.sum` to the e2e folder (all Golang stuff should be in the e2e folder imo)
* Delete `node_modules` and `yarn.lock` in the root directory (someone probably ran `yarn install` here by mistake)
* Run formatter for Makefile


larry0x:
I'd also like to do another PR to update versioning and dependency management using [this approach](https://twitter.com/larry0x/status/1582200990060941312) after this one.

jhernandezb (approved):
thank you! looks good to me.

PR #438, opened 2022-10-18
Use cargo workspace feature to manage versions and dependencies

larry0x:
See: https://twitter.com/larry0x/status/1582200990060941312

The benefit of this approach is that, say you have 20 crates in the repo, and they all depend on `cosmwasm-std`. To bump of the version of `cosmwasm-std`, you won't need to go into the `Cargo.toml` of each crate and edit them individually. Instead you just edit it once in the workspace-level.

Notes:

- **This PR makes all crates in this repo (both contracts and packages) to have the workspace-level version of v0.20.0.** For example and `base-factory` contracts jumps directly form `0.2.0` to `0.20.0`. Let me know if this is desired. It is possible to set individual crate versions different from the workspace-level version.
- I applied some styling based on my personal preference in the dependencies lists (specifically, order all deps alphabetically and align the equation marks). Let me know if this is ok; can revert.
- An implication of this PR is that a minimum Rust version of 1.64.0 will be required for working this repo. Have to bump the docker container versions used in CI due to this reason.


larry0x:
Currently CI fails also due to it using Rust 1.63.

I did build the contracts and ran tests (unit, integration, e2e) locally and they pass.

shanev (approved):
Thanks so much for this @larry0x! Wanted to do this after seeing your tweet so very grateful for this. Will merge it in and fix CI.

PR #450, opened 2022-10-25
add transaction index to the seed

jhernandezb:
Previous version of CosmWasm was not providing the right tx index from the environment, this has been fixed in 0.29 and cosmwasm 1.1.

This PRs adds back the tx index which will depend on how many txs land on the same block.


yubrew (commented):
👍

PR #469, opened 2022-11-18
add discount price

jhernandezb:


contracts/minter/src/msg.rs
```diff
@@ -26,6 +26,7 @@ pub enum ExecuteMsg {
     MintTo { recipient: String },
     MintFor { token_id: u32, recipient: String },
     UpdateUnitPrice { price: u128 },
+    UpdateDiscountPrice { price: u128 },
```
shanev, line 29: I would make this a `Uint128` so it can be presented as a string by the client. May run into big int issues on the frontend using a `u128`.

shanev (changes requested):

shanev (approved):

PR #479, opened 2022-12-08
Refactor Vending Minter Tests

humanalgorithm:
PART 1
There are 3 PRs in this series: 
1) https://github.com/public-awesome/launchpad/pull/479  => Refactors the vending minter tests that will serve as the common core for other similar tests
2) https://github.com/public-awesome/launchpad/pull/483  => Create the test-suite folder where all tests will eventually go and move the two contracts affected most to the test-suite
3) https://github.com/public-awesome/launchpad/pull/486  => Shape and move all of the remaining tests to the test-suite folder

Motivation: In the sg-eth-airdrop PR I had to borrow a good amount of the code for the vending minter tests.  Since I did a refactor there, figure I would continue on to the source and do some refactoring here. This idea later evolved into sharing all of the commonly reused code for a test-suite folder (developed in parts 2 and 3). 
 
 1) The main feature added here is the `configure_minter` function. The interface of the function looks like so: 
 ```rs
 pub fn configure_minter(
    app: &mut StargazeApp,
    minter_admin: Addr,
    collection_params_vec: Vec<CollectionParams>,
    minter_instantiate_params_vec: Vec<MinterInstantiateParams>,
) -> Vec<MinterCollectionResponse> {
```
2) configure_minter takes two vectors, 1) a vector of collection parameters for providing whichever collections should be tied to minters and 2) a vector of MinterInstantiateParams which contain some settings when instantiating a minter (such as minter start time or splits address). Since vectors are provided, a dev could instantiate any number of minter contracts tied to collections (this may be useful in marketplace if we want to quickly instantiate a whole list of collections). 
3) `configure_minter` returns a vector of `MinterCollectionResponse` objects. The message looks like so: 
```rs
pub struct MinterCollectionResponse {
    pub minter: Option<Addr>,
    pub collection: Option<Addr>,
    pub factory: Option<Addr>,
    pub error: Option<Error>,
}
```
In this way, a developer can call configure minter and get back the minter, factory and collection addresses for each one that is generated in the test. 
4) The error return of `MinterCollectionResponse` allows a dev to peer into any errors returned quickly without special logic. In the following example we try to generate a collection before genesis time and get an error:
```rs
    let minter_collection_response: Vec<MinterCollectionResponse> = configure_minter(
        &mut router,
        creator.clone(),
        vec![collection_params.clone()],
        vec![minter_params],
    );

    let err = minter_collection_response[0]
        .error
        .as_ref()
        .unwrap()
        .root_cause();
    let expected_error = ContractError::BeforeGenesisTime {};
 ```


shanev (approved):
🙏🙏🙏

PR #481, opened 2022-12-13
Add executor to splits

shanev:
An executor can be a member of the group, any address, or `None`. A zero weight member can also be an executor. This is useful for bots. i.e: A bot tied to a cron job that calls `distribute()` periodically.


contracts/splits/src/contract.rs
```diff
@@ -35,7 +35,10 @@ pub fn instantiate(
 
     set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
 
-    let cfg = Config { group_addr };
+    let cfg = Config {
+        group_addr,
+        executor: msg.executor,
```
jhernandezb, line 40: address is not being validated if it is Only(Addr)

contracts/splits/src/contract.rs
```diff
@@ -60,16 +63,15 @@ pub fn execute_distribute(
...
 
-    // only a member can distribute funds
-    let weight = config
-        .group_addr
-        .is_member(&deps.querier, &info.sender, None)?
-        .ok_or(ContractError::Unauthorized {})?;
-    if weight == 0 {
-        return Err(ContractError::InvalidWeight { weight });
-    }
+    config.authorize(&deps.querier, &info.sender)?;
```
jhernandezb, line 66: I think would be nice to allow to be either an executor or any of the members, but seems like with this change you have to stick with one method, I guess can always set a bot to have 0 weight.

contracts/splits/src/contract.rs
yubrew: we should set sane defaults, and i think the default should be member-gated distribute instead of anyone can distribute.

yubrew (commented):
i like separating of executor as a separate role, but how cw4 is set up, i think it makes more sense for a distribute bot to be added as 0 weight member, and any member can distribute.

if there is additional future functionality separating members and bot functionality, we can add back the executor role.

contracts/splits/src/contract.rs
shanev: Yes a 0 weight member makes more sense for the bot imho.

PR #485, opened 2022-12-16
Update splits executor via gov or current executor

shanev:


shanev:
Replaced by admin controller.

PR #492, opened 2023-01-14
[SG-203] Allow creating splits with an existing group

shanev:
Splits can now be created by passing in the address of a group, or it can create it's own group given the group's `code_id`. This makes it more flexible, and provides a better UX for frontend for easily creating splits.

Also introduces a `init` controller, some convenient code for use when a contract instantiates another.


contracts/splits/src/contract.rs
```diff
@@ -1,51 +1,55 @@
...
-    }
-
-    let cfg = Config {
-        group_addr,
-        executor: msg.executor,
-    };
-    CONFIG.save(deps.storage, &cfg)?;
+    let self_addr = env.contract.address;
+
+    match msg.group {
```
humanalgorithm, line 33: I think you can put this in a function that is that is just called "group_message( )?" or something like that.

contracts/splits/src/contract.rs
```diff
@@ -1,51 +1,55 @@
...
+        )),
+        Group::Cw4Address(addr) => {
+            let group = Cw4Contract(
+                deps.api
+                    .addr_validate(&addr)
+                    .map_err(|_| ContractError::InvalidGroup { addr })?,
+            );
+
+            let weight = group.total_weight(&deps.querier)?;
+            if weight == 0 {
```
humanalgorithm, line 45: can we do a "weight_equals_zero_check()?" Seeing this line a lot.

packages/controllers/src/init.rs
```diff
@@ -0,0 +1,31 @@
...
+#[cw_serde]
+pub struct ContractInstantiateMsg {
+    pub code_id: u64,
+    pub msg: Binary,
+    pub admin: Option<Admin>,
+    pub label: String,
+}
+
+impl ContractInstantiateMsg {
+    pub fn into_wasm_msg(self, creator: Addr) -> WasmMsg {
```
humanalgorithm, line 19: Nice 👍

contracts/splits/src/contract.rs
shanev: Overkill imho.

PR #497, opened 2023-01-19
SG-266 add more splits tests

yubrew:


contracts/splits/src/contract.rs
```diff
@@ -84,6 +87,7 @@ pub fn execute_distribute(
     let group = GROUP.load(deps.storage)?;
 
     let total_weight = checked_total_weight(&group, deps)?;
+    checked_total_members(&group, deps)?;
```
jhernandezb, line 90: we are querying twice, one for checking total members and for send msgs, I'd like to avoid the extra query to make it a bit more gas efficient.

yubrew: resolved

shanev (commented):
Use 256-bit math for better precision in distribution.

shanev (approved):

shanev:
> Use 256-bit math for better precision in distribution.

Normalized distribution math is fine. Precision isn't too much of an issue with large weights.

jhernandezb (approved):

PR #506, opened 2023-02-08
Increase dev reward to 50%

shanev:


jhernandezb:
This increase to 50% seems like it's more like a 25% because is taken from the initial half amount.

shanev:
Make it 50% burn, 50% dev?

jhernandezb:
seems like prop passed 50% should go to dev and the rest fairburned

shanev:
Closing this because we want to have all gov params in a smart contract.

shanev:
Update to burn 50% instead of going to stakers.

packages/sg1/src/lib.rs
```diff
@@ -102,12 +103,12 @@ mod tests {
         fair_burn(1000u128, Some(Addr::unchecked("geordi")), &mut res);
         let bank_msg = SubMsg::new(BankMsg::Send {
             to_address: "geordi".to_string(),
-            amount: coins(100, "ustars".to_string()),
+            amount: coins(500, NATIVE_DENOM),
```
humanalgorithm, line 106: Should we have a test where the amount for rewards is not an even number (to test the decimal placement)?

humanalgorithm (approved):

packages/sg1/src/lib.rs
```diff
@@ -40,10 +40,7 @@ pub fn fair_burn(fee: u128, developer: Option<Addr>, res: &mut Response) {
             }));
             event = event.add_attribute("dev", dev.to_string());
             event = event.add_attribute("dev_amount", Uint128::from(dev_fee).to_string());
-            (
-                Decimal::percent(FEE_BURN_PERCENT - DEV_INCENTIVE_PERCENT),
-                dev_fee,
-            )
+            (Decimal::percent(FEE_BURN_PERCENT), dev_fee)
```
tasiov, line 43: you could pull `FEE_BURN_PERCENT` out of this `match developer` block since it is not used there

tasiov (commented):
I think you will hit a rounding error on odd numbers with the current implementation.

DEV_PERCENT = 50%
FAIR_BURN_PERCENT = 50%

9 STARS fee
9 * .5 = 4 dev fee
9 * .5 = 4 fair burn fee
1 remaining

Maybe it would make sense to calculate the fair burn fee, and then pay the remainder to the developer or the community pool.

PR #510, opened 2023-02-27
Fix base factory deploy script

shanev:


jhernandezb (approved):

.circleci/config.yml
```diff
@@ -492,7 +492,7 @@ jobs:
       - run:
           name: Build development contracts
           command: |
-            docker run --volumes-from with_code cosmwasm/workspace-optimizer:0.12.11
+            docker run --volumes-from with_code cosmwasm/workspace-optimizer:0.12.9
```
jhernandezb, line 495:
```suggestion
            docker run --volumes-from with_code cosmwasm/workspace-optimizer:0.12.11
```

jhernandezb (approved):

.circleci/config.yml
jhernandezb: this is fine to bump, the issue was with the docker -arm  version in optimize.sh which breaks CI.

scripts/optimize.sh
```diff
@@ -1,4 +1,4 @@
-docker run --rm -v "$(pwd)":/code \
-    --mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
-    --mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
-    cosmwasm/workspace-optimizer-arm64:0.12.11
\ No newline at end of file
+docker run --rm -v "$(pwd)":/code --platform linux/amd64 \
+	--mount type=volume,source="$(basename "$(pwd)")_cache",target=/code/target \
+	--mount type=volume,source=registry_cache,target=/usr/local/cargo/registry \
+	cosmwasm/workspace-optimizer:0.12.11
```
jhernandezb, line 4: perhaps this script can be improved later to detect ARM  hosts (m1/m2) and run with different docker image instead.

PR #531, opened 2023-03-27
SG-619 Add sudo freeze

yubrew:
add sudo msg to stop factory from creating new minters


shanev (changes requested):
So old factories will have to be migrated to this in order to be frozen right? So this could use a migrate function that adds the `frozen` field and sets it to true.

May be excessive but here's an example of testing a migration: https://github.com/DA0-DA0/dao-contracts/blob/b09db5a672204229fdd4722c9294b9dd25f4ed03/contracts/proposal/dao-proposal-single/src/testing/migration_tests.rs#L35.

jhernandezb:
Our marketplace also has an example

https://github.com/public-awesome/marketplace/blob/v1.0.0/contracts/marketplace/src/execute.rs#L1282

Which is more simpler

jhernandezb:
To speed up this deployment perhaps we can just freeze old code ids for vending-minter and sg721-base

jhernandezb (approved):

contracts/base-factory/src/contract.rs
```diff
@@ -135,6 +141,10 @@ pub fn update_params<T, C>(
         params.min_mint_price = min_mint_price;
     }
 
+    if let Some(frozen) = param_msg.frozen {
+        params.frozen = frozen;
+    }
+
```
jhernandezb, line 147: move to the top

jhernandezb (approved):

shanev (approved):

PR #532, opened 2023-03-28
Move contracts into folders

shanev:
Organizes contracts better. Now it's apparent that new types of factories, minters, collections, and whitelists can be built. For example, a new open edition minter can go in `minters`. A new Merkle whitelist can go in `whitelists`.


jhernandezb:
scripts/release-* might need an update to ensure deterministic release of crates.

shanev:
> scripts/release-* might need an update to ensure deterministic release of crates.

They were updated as part of this PR.

jhernandezb (approved):

PR #535, opened 2023-03-31
sg721 updatable v1

jhernandezb:


contracts/sg721-updatable/src/contract.rs
```diff
@@ -0,0 +1,367 @@
...
+use crate::state::{CollectionInfo, RoyaltyInfo, COLLECTION_INFO, FROZEN_TOKEN_METADATA};
+
+// version info for migration info
+const COMPATIBLE_MIGRATION_CONTRACT_NAME: &str = "crates.io:sg-721";
+const CONTRACT_NAME: &str = "crates.io:sg721-updatable";
+const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+
+const CREATION_FEE: u128 = 1_000_000_000;
+
+// Disable instantiation for production build
```
shanev, line 32: Don't need this anymore?

shanev (changes requested):

contracts/sg721-updatable/src/contract.rs
jhernandezb: we need it I disabled because I though it was something wrong with my tests but should be added back

contracts/sg721-updatable/src/contract.rs
```diff
@@ -0,0 +1,367 @@
...
+        .tokens
+        .update(deps.storage, &token_id, |token| match token {
+            Some(mut token_info) => {
+                token_info.token_uri = token_uri.clone();
+                Ok(token_info)
+            }
+            None => Err(ContractError::TokenIdNotFound {}),
+        })?;
+
+    let event = Event::new("update_update_token_metadata")
```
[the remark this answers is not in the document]
yubrew: yes, it's the same event

PR #539, opened 2023-04-01
Update cw-nfts version

humanalgorithm:
Update 4-24-2023:  This is now using the inherited migrations from cw_721 base per larry engineer PR: https://github.com/CosmWasm/cw-nfts/pull/119

Updating to the latest cw-nfts 0.17 and the necessary changes. 

Notes: 
1) In order to update to cw 0.17 need to update the Rust version to 1.67.1. You will see this reflected in the changes to the config.yml build script. 

2) The MintMsg has been removed and folded into ExecuteMsg for cw721-base. Therefore we now need to put the variant for Mint { } in each relevant msg module. 

3) Need to use a NftParams enum for nft data passed to mint function. This is such as to not have too many parameters passed to the function. Would use a regular struct but need to be able to plug into the T extension parameter.


Cargo.toml
```diff
@@ -22,15 +22,15 @@ base-factory        = { version = "2.1.0", path = "contracts/factories/base-fact
...
-cw2                 = "0.16.0"
-cw4                 = "0.16.0"
-cw4-group           = "0.16.0"
-cw721               = "0.16.0"
-cw721-base          = "0.16.0"
-cw-multi-test       = "0.16.0"
-cw-storage-plus     = "0.16.0"
-cw-utils            = "0.16.0"
+cw-controllers      = "1.0.1"
+cw2                 = { git = "https://github.com/CosmWasm/cw-plus/", branch="main" }
```
shanev, line 26: Why isn't this also 1.0.1?

contracts/collections/sg721-base/src/contract.rs
```diff
@@ -54,7 +56,8 @@ where
         self.parent.contract_info.save(deps.storage, &info)?;
 
         let minter = deps.api.addr_validate(&msg.minter)?;
-        self.parent.minter.save(deps.storage, &minter)?;
+        self.minter.save(deps.storage, &minter)?;
+        cw_ownable::initialize_owner(deps.storage, deps.api, Some(&msg.minter))?;
```
shanev, line 60: So now the minter it stored in two places which is redundant. The owner should be the creator, but is already in `CollectionInfo`. I understand setting this is necessary per cw721 changes, but maybe we can refactor sg721 to include cw_ownable and remove redundancy. 

@larry0x Have any ideas here?

contracts/collections/sg721-base/src/contract.rs
```diff
@@ -335,6 +372,20 @@ where
             royalty_info: royalty_info_res,
         })
     }
+
+    pub fn migrate(deps: DepsMut, _env: Env, _msg: Empty) -> Result<Response, ContractError> {
+        // make sure the correct contract is being upgraded, and it's being
+        // upgraded from the correct version.
+        let version = cw2::get_contract_version(deps.storage)?;
+        cw2::assert_contract_version(deps.as_ref().storage, CONTRACT_NAME, EXPECTED_FROM_VERSION)
```
shanev, line 380: I think checking the exact previous version is excessive because it adds more work for every migration and only allows migrating from the previous version to the next. Just making sure the new version is greater than the old version is fine imho.

contracts/collections/sg721-base/src/error.rs
```diff
@@ -46,4 +46,13 @@ pub enum ContractError {
...
     #[error("RoyaltyShareIncreased")]
     RoyaltyShareIncreased {},
+
+    #[error("MinterNotFound")]
+    MinterNotFound {},
+
+    #[error("Ownership Update Error: {error}")]
+    OwnershipUpdateError { error: String },
+
+    #[error("found version ({0}) while attempting to migrate from 0.16.0")]
```
shanev, line 56: Hardcoding 0.16.0 here means that a developer has to remember it has to be updated every time.

contracts/collections/sg721-base/src/lib.rs
```diff
@@ -24,8 +25,9 @@ pub mod entry {
     use sg_std::Response;
 
     // version info for migration info
-    const CONTRACT_NAME: &str = "crates.io:sg721-base";
-    const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+    pub const CONTRACT_NAME: &str = "crates.io:sg721-base";
+    pub const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+    pub const EXPECTED_FROM_VERSION: &str = "0.16.0";
```
shanev, line 30: I don't think old versions should be hardcoded. Extra work for developers and more things to remember.

contracts/collections/sg721-base/src/msg.rs
```diff
@@ -199,3 +232,13 @@ impl CollectionInfoResponse {
         }
     }
 }
+
+#[cw_serde]
+pub enum NftParams<T> {
```
shanev, line 237: Can't this just be a struct?

contracts/collections/sg721-base/src/upgrades/v0_17.rs
```diff
@@ -0,0 +1,28 @@
...
+    let tract16 = v16::Cw721Contract::<T, C, E, Q>::default();
+    let minter = tract16.minter.load(deps.storage)?;
+    tract16.minter.remove(deps.storage);
+
+    // save new ownership info
+    let ownership = cw_ownable::initialize_owner(deps.storage, deps.api, Some(minter.as_str()))?;
+
+    Ok(Response::new()
+        .add_attribute("action", "migrate")
+        .add_attribute("from_version", "0.16.0")
```
shanev, line 24: Hardcoding versions make upgrading cumbersome. Should be read from env vars.

contracts/collections/sg721-metadata-onchain/src/lib.rs
```diff
@@ -3,15 +3,18 @@ use cosmwasm_std::Empty;
...
-
 pub type Sg721MetadataContract<'a> = sg721_base::Sg721Contract<'a, Metadata>;
 pub type InstantiateMsg = sg721::InstantiateMsg;
 pub type ExecuteMsg = sg721::ExecuteMsg<Metadata, Empty>;
 pub type QueryMsg = sg721_base::msg::QueryMsg;
 
+// version info for migration info
+const CONTRACT_NAME: &str = "crates.io:sg721-metadata-onchain";
+const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+pub const EXPECTED_FROM_VERSION: &str = "0.16.0";
```
shanev, line 14: Same as before.

contracts/collections/sg721-metadata-onchain/src/lib.rs
```diff
@@ -50,6 +53,21 @@ pub mod entry {
     pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
         Sg721MetadataContract::default().query(deps, env, msg)
     }
+
+    #[entry_point]
+    pub fn migrate(deps: DepsMut, _env: Env, _msg: Empty) -> Result<Response, ContractError> {
+        // make sure the correct contract is being upgraded, and it's being
+        // upgraded from the correct version.
+        let version = cw2::get_contract_version(deps.storage)?;
+        cw2::assert_contract_version(deps.as_ref().storage, CONTRACT_NAME, EXPECTED_FROM_VERSION)
```
shanev, line 62: Same as before.

contracts/collections/sg721-nt/src/lib.rs
```diff
@@ -1,23 +1,27 @@
...
 use sg721_base::Sg721Contract;
 pub type QueryMsg = sg721_base::msg::QueryMsg;
 pub type Sg721NonTransferableContract<'a> = Sg721Contract<'a, Extension>;
+use sg721_base::msg::NftParams;
+
+// version info for migration info
+// version info for migration info
+const CONTRACT_NAME: &str = "crates.io:sg721-nt";
+const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
+pub const EXPECTED_FROM_VERSION: &str = "0.16.0";
```
shanev, line 16: Same as before.

contracts/collections/sg721-nt/src/lib.rs
```diff
@@ -71,4 +88,19 @@ pub mod entry {
     pub fn query(deps: Deps, env: Env, msg: QueryMsg) -> StdResult<Binary> {
         Sg721NonTransferableContract::default().query(deps, env, msg)
     }
+
+    #[entry_point]
+    pub fn migrate(deps: DepsMut, _env: Env, _msg: Empty) -> Result<Response, ContractError> {
+        // make sure the correct contract is being upgraded, and it's being
+        // upgraded from the correct version.
+        let version = cw2::get_contract_version(deps.storage)?;
+        cw2::assert_contract_version(deps.as_ref().storage, CONTRACT_NAME, EXPECTED_FROM_VERSION)
```
shanev, line 97: Same as before.

contracts/collections/sg721-updatable/src/contract.rs
```diff
@@ -31,6 +35,21 @@ pub fn _instantiate(
     Ok(res)
 }
 
+#[entry_point]
+pub fn migrate(deps: DepsMut, _env: Env, _msg: Empty) -> Result<Response, sg721BaseContractError> {
+    // make sure the correct contract is being upgraded, and it's being
+    // upgraded from the correct version.
+    let version = cw2::get_contract_version(deps.storage)?;
+    cw2::assert_contract_version(deps.as_ref().storage, CONTRACT_NAME, EXPECTED_FROM_VERSION)
```
shanev, line 43: Same as before.

shanev (changes requested):
Have to re-think ownership and hardcoding upgrade versions.

contracts/collections/sg721-base/src/lib.rs
humanalgorithm: Note that I am following the logic from the cw 0.17 upgrade here: 
https://github.com/CosmWasm/cw-nfts/pull/112/files#diff-733b03d19162a7149fd3e3feb6552118ea6a667aac4cbcda9ece681ad6425ef7R39

Please refer to this for all of the 0.16 comments. It wouldn't make sense for me to do the update logic in a different way than the underlying cw-nfts logic that I am copying.

contracts/collections/sg721-metadata-onchain/src/lib.rs
humanalgorithm: Again, am copying from cw-nfts migrate logic: 
https://github.com/CosmWasm/cw-nfts/pull/112/files#diff-733b03d19162a7149fd3e3feb6552118ea6a667aac4cbcda9ece681ad6425ef7R83

contracts/collections/sg721-base/src/contract.rs
humanalgorithm: Copying logic from cw-nfts: https://github.com/CosmWasm/cw-nfts/pull/112/files#diff-733b03d19162a7149fd3e3feb6552118ea6a667aac4cbcda9ece681ad6425ef7R83

Cargo.toml
humanalgorithm: See the logic from cw-nfts, they wanted to use a non-main commit in order to get at the assert_contract_version function (note "main" is not part of the 1.0.1 yet):  https://github.com/CosmWasm/cw-nfts/pull/112/files#diff-2e9d962a08321605940b5a657135052fbcef87b5e360662bb527c96d9a615542R18

contracts/collections/sg721-base/src/error.rs
yubrew: The error I expect is "migration version error" trying to migrate from v1 to v2 for example as commented in `contract.rs`.

contracts/collections/sg721-base/src/lib.rs
yubrew: The other way to do this is a set of whitelisted versions. See sg721-updatable/contract.rs migrate fn

contracts/collections/sg721-base/src/msg.rs
```diff
@@ -46,46 +66,58 @@ pub enum ExecuteMsg<T, E> {
...
         limit: Option<u32>,
     },
+    #[returns(NumTokensResponse)]
     NumTokens {},
+    #[returns(ContractInfoResponse)]
     ContractInfo {},
-    NftInfo {
-        token_id: String,
-    },
+    #[returns(NftInfoResponse<Empty>)]
```
yubrew, line 100: double check this works for sg721-metadata-onchain

contracts/collections/sg721-metadata-onchain/Cargo.toml
```diff
@@ -38,7 +38,10 @@ sg721           = { workspace = true }
 sg721-base      = { workspace = true, features = ["library"] }
 sg-metadata     = { workspace = true }
 sg-std          = { workspace = true }
+cw721-base-016  = { workspace = true }
```
yubrew, line 41: is this necessary? can you just add this to tests?

contracts/collections/sg721-base/src/contract.rs
humanalgorithm: @yubrew are you asking to change something in the code here? Not sure what the action to take is.

contracts/collections/sg721-base/src/error.rs
humanalgorithm: Going to use this so that error object is not hard coded but not that we are still using the expected from constant: 

```rs
    cw2::assert_contract_version(deps.as_ref().storage, CONTRACT_NAME, EXPECTED_FROM_VERSION)
        .map_err(|_| {
            sg721_base::ContractError::WrongMigrateVersion(
                version.version,
                EXPECTED_FROM_VERSION.to_string(),
            )
        })?;
```

contracts/collections/sg721-base/src/msg.rs
humanalgorithm: How to double check? Do you mean after merge?

contracts/collections/sg721-metadata-onchain/Cargo.toml
humanalgorithm: Seems like this is a redundant package now.

contracts/collections/sg721-base/src/lib.rs
humanalgorithm: @yubrew can you clarify? I don't see the word "migrate" in this file: 

https://github.com/public-awesome/launchpad/blob/main/contracts/collections/sg721-updatable/src/contract.rs

Cargo.toml
larry0x: We need Confio to make a new release of cw2 that includes `assert_contract_version`

contracts/collections/sg721-base/src/contract.rs
larry0x: > now the minter it stored in two places

Actually it's stored in 3 places?

1. in `CollectionInfo` as the creator
2. in `self.minter`
3. in `cw_ownable::OWNERSHIP`

Imo, (1) is ok - I consider the creator a separate concept from the minter. Consider a physical painting. The creator is the painter who painted it. Then there's the current owner of the painting which can be a separate person. Although in sg721 the creator and minter may always be the same but we can imagine in the future they may be allowed to be different accounts.

(2) - not sure how sg721 makes use of `self.minter`. If it uses cw-ownable to define who has minting power then `self.minter` may be redundant (and can be deleted during the migration).

contracts/collections/sg721-base/src/contract.rs
larry0x: > Just making sure the new version is greater than the old version is fine imho.

@shanev If the migration from version `A` to version `B` includes some state changes, and from `B` to `C` also includes some state changes, then you can't directly apply the version `C` migration method on a contract that's still on `A`.

contracts/collections/sg721-base/src/lib.rs
larry0x: > I don't think old versions should be hardcoded.

@shanev I'm considering organizing upgrade logics like this:

```plain
.
└── src/
    └── upgrades/
        ├── mod.rs
        ├── v0_17.rs
        ├── v0_18.rs
        └── v0_19.rs
```

`v0_17.rs` contains the migration logics from 0.16 to 0.17; `v0_18.rs` contains logics for 0.17 => 0.18, so on.

In 0.16 => 0.17 we may need to assert that the contract being upgraded is at 0.16. This may be because older versions (e.g. 0.15) may have different layouts that are incompatible with the migration logics. In this case, imo it's reasonable to hardcode a version.

Similarly if 0.17 => 0.18 migration logics only work on 0.17, it may also hardcode an expected version accordingly.

contracts/collections/sg721-base/src/upgrades/v0_17.rs
larry0x: As I suggested in earlier comments, `v0_17.rs` contains the specific upgrade logics for 0.16 => 0.17 and not any other migrations, therefore imo it's reasonable to hardcode it here.

Cargo.toml
```diff
@@ -22,15 +22,15 @@ base-factory        = { version = "2.3.0", path = "contracts/factories/base-fact
...
-cw721               = "0.16.0"
-cw721-base          = "0.16.0"
-cw-multi-test       = "0.16.0"
-cw-storage-plus     = "0.16.0"
-cw-utils            = "0.16.0"
+cw-controllers      = "1.0.1"
+cw2 = "1.0.1"
+cw4                 = "1.0.1"
+cw4-group           = "1.0.1"
+cw721               = { git = "https://github.com/CosmWasm/cw-nfts/", branch="main" }
```
shanev, line 29: Is the latest version not published? If so I can do that.

contracts/collections/sg721-base/src/state.rs
```diff
@@ -11,6 +11,7 @@ where
     T: Serialize + DeserializeOwned + Clone,
 {
     pub parent: Parent<'a, T>,
+    pub minter: Item<'a, Addr>,
```
jhernandezb, line 14: We shouldn't be adding this field, instead check against cw_owner

contracts/collections/sg721-base/src/contract.rs
```diff
@@ -54,7 +56,8 @@ where
         self.parent.contract_info.save(deps.storage, &info)?;
 
         let minter = deps.api.addr_validate(&msg.minter)?;
-        self.parent.minter.save(deps.storage, &minter)?;
+        self.minter.save(deps.storage, &minter)?;
```
jhernandezb, line 59: this will need to go away and use cw_owner only.

contracts/collections/sg721-base/src/contract.rs
```diff
@@ -240,7 +269,7 @@ where
         info: MessageInfo,
         start_time: Option<Timestamp>,
     ) -> Result<Response, ContractError> {
-        let minter = self.parent.minter.load(deps.storage)?;
+        let minter = self.minter.load(deps.storage)?;
         if minter != info.sender {
             return Err(ContractError::Unauthorized {});
         }
```
jhernandezb, line 275: this needs to check cw_owner instead of self.minter

jhernandezb (changes requested):

contracts/collections/sg721-base/src/contract.rs
humanalgorithm: This has been addressed by using ```        let owner = cw_ownable::get_ownership(deps.storage);```

contracts/collections/sg721-base/src/contract.rs
humanalgorithm: This has been addressed by using    ```let owner = cw_ownable::get_ownership(deps.storage);```

shanev (approved):

yubrew:
Updated title so linear catches issue

PR #540, opened 2023-04-01
Enable minting with any IBC asset

shanev:
* Factory allows a minimum mint price in any denom
* Non-native fees are sent to the community pool


contracts/factories/vending-factory/src/contract.rs
```diff
@@ -83,9 +83,10 @@ pub fn execute_create_minter(
         });
     }
 
-    if NATIVE_DENOM != msg.init_msg.mint_price.denom {
-        return Err(ContractError::BaseError(BaseContractError::InvalidDenom {}));
-    }
+    ensure!(
+        params.min_mint_price.denom == msg.init_msg.mint_price.denom,
+        ContractError::DenomMismatch {}
+    );
```
jhernandezb, line 89: This seems to not be covered by tests

contracts/minters/vending-minter/src/contract.rs
```diff
@@ -690,9 +701,15 @@ fn _execute_mint(
         .add_attribute("sender", info.sender)
         .add_attribute("recipient", recipient_addr)
         .add_attribute("token_id", mintable_token_mapping.token_id.to_string())
-        .add_attribute("network_fee", network_fee)
-        .add_attribute("mint_price", mint_price.amount)
-        .add_attribute("seller_amount", seller_amount))
+        .add_attribute(
+            "network_fee",
+            coin(network_fee.u128(), mint_price.clone().denom).to_string(),
+        )
```
jhernandezb, line 707: I don't think indexers use this but this will probably break it since going from number to amount+denom

contracts/minters/vending-minter/src/contract.rs
```diff
@@ -690,9 +701,15 @@ fn _execute_mint(
...
+        .add_attribute("mint_price", mint_price.to_string())
```
jhernandezb, line 708: same for mint_price

jhernandezb (changes requested):
Couple comments
Missing
- Allow whitelists to allow any denom as well
- Vending minter needs to check whitelist price is within the same allowed denom

contracts/minters/vending-minter/src/contract.rs
shanev: Right, indexer will have to be updated to parse full coins for this new version and beyond.

contracts/minters/vending-minter/src/contract.rs
shanev: Same, have to update indexer -- will make a new bug.

PR #545, opened 2023-04-06
[WIP] Fix metamask ledger issue

humanalgorithm:
The sg-eth-airdrop does not work when someone uses a ledger with metamask to sign a transaction and send to the validation methods. 

I ran the metamask ledger path and then copy/pasted my address and signature into a unit test. 
That unit test in this PR is the only way to demonstrate the failure other than someone running a metamask with ledger end to end. 
If we can get this test to pass then we will know that the fix is implemented. 

Fix should rewrite the file at packages/ethereum-verify/src/signature_verify.rs


shanev (approved):
Mind adding some comments / docs into the backstory here?

PR #562, opened 2023-04-25
Fix typo in tokens_response, all_tokens_response schema

jinmel:
- Rename start_from to start_after following the actual contract definition.


yubrew:
I'm going to close this PR for now since it should really be fixed in [cw-nfts](https://github.com/CosmWasm/cw-nfts)

PR #572, opened 2023-05-24
add whitelist mint count to flex minter

jhernandezb:
for vending-minter-flex we use different counters for whitelist and for public mint, however there is no way to retrieve such information without querying raw keys, this change adds the whitelist counter to the query response.


shanev (changes requested):
Want to add a test?

shanev (approved):

PR #575, opened 2023-06-05
Replace `admin` with cw-ownable for Vending Minter Flex

shanev:
This replaces `admin` and replaces it with cw-ownable.

Note this change is on vending-minter-flex as it appears to be the most up-to-date. We should probably move the old vending minter tests to this flex one, and delete vending-minter.

Since there were no tests in vending-minter-flex, this PR doesn't include any tests as well. When tests are migrated over, then they should be updated to handle cw-ownable.


jhernandezb:
Closing for now since it needs more design considerations for sg721 conflicting with `cw_ownable (minter)` and `collection_info.creator`

PR #577, opened 2023-06-08
Royalty increase and add `AddressLike`

shanev:


yubrew (approved):

packages/sg721/src/lib.rs
```diff
@@ -92,31 +95,26 @@ pub struct UpdateCollectionInfoMsg<T> {
 }
 
 #[cw_serde]
-pub struct RoyaltyInfo {
-    pub payment_address: Addr,
+pub struct RoyaltyInfo<T: AddressLike> {
+    pub payment_address: T,
     pub share: Decimal,
+    pub updated_at: Timestamp,
```
jhernandezb, line 101: This will probably break de-serialization if a contract is migrating to this new version

packages/sg721/src/lib.rs
```diff
@@ -92,31 +95,26 @@ pub struct UpdateCollectionInfoMsg<T> {
 }
 
 #[cw_serde]
-pub struct RoyaltyInfo {
-    pub payment_address: Addr,
+pub struct RoyaltyInfo<T: AddressLike> {
+    pub payment_address: T,
```
jhernandezb, line 99: How is Addr stored? as string? I guess is fully compatible .

contracts/collections/sg721-base/src/contract.rs
```diff
@@ -225,27 +223,63 @@ where
...
+
+                // if royalty share changed and current time is after last royalty update + min duration
+                // royalty_updated_at is always Some because it is set in instantiate
+                if (new_royalty_info_res.share != curr_royalty_info_res.share) && update_too_soon {
+                    return Err(ContractError::RoyaltyUpdateTooSoon {});
                 }
             } else {
-                return Err(ContractError::RoyaltyShareIncreased {});
+                return Err(ContractError::RoyaltyShareIncreasedTooMuch {});
             }
```
jhernandezb, line 274: This here doesn't allow to increase it if it was empty before. Perhaps you want to allow it now with the same time interval.

contracts/collections/sg721-base/src/contract.rs
```diff
@@ -225,27 +223,63 @@ where
...
+
+                // make sure the update time is after 24 hours
+                let end = (Expiration::AtTime(curr_royalty_info_res.updated_at) + DAY)?;
+                let update_too_soon = !end.is_expired(&env.block);
+
+                // if royalty share changed and current time is after last royalty update + min duration
+                // royalty_updated_at is always Some because it is set in instantiate
+                if (new_royalty_info_res.share != curr_royalty_info_res.share) && update_too_soon {
+                    return Err(ContractError::RoyaltyUpdateTooSoon {});
                 }
```
jhernandezb, line 271: Do you want to apply same logic for decreasing royalties? I guess you can only decrease it  once every 24 hours.

contracts/collections/sg721-base/src/contract.rs
```diff
@@ -225,27 +223,63 @@ where
...
+            .query_wasm_smart(minter_address, &MinterQueryMsg::Config {})?;
+        let factory_address = minter_config.factory;
+        let factory_params: ParamsResponse<T> = deps
+            .querier
+            .query_wasm_smart(factory_address, &Sg2QueryMsg::Params {})?;
+        let max_royalty = factory_params.params.max_royalty_bps.bps_to_decimal();
+        let max_royalty_increase_rate = factory_params
+            .params
+            .max_royalty_increase_rate_bps
+            .bps_to_decimal();
```
jhernandezb, line 246: So this means we will have to migrate the factory, have we verified this won't break previous minters?

packages/sg721/src/lib.rs
```diff
@@ -92,31 +95,26 @@ pub struct UpdateCollectionInfoMsg<T> {
 }
 
 #[cw_serde]
-pub struct RoyaltyInfo {
-    pub payment_address: Addr,
+pub struct RoyaltyInfo<T: AddressLike> {
```
jhernandezb, line 98: This struct is also used by the marketplace, not sure what happens to serde when there is unexpected new fields.

jhernandezb (changes requested):

jhernandezb:
- Test factory migration do not break minters previously created
- Test migrating sg721-base doesn't break Marketplace when it queries for royalties and start_trading_time
- Test you can increase royalties and still list/buy/sell

packages/sg721/src/lib.rs
yubrew: fixed w Option<Timestamp>

packages/sg721/src/lib.rs
yubrew: Think Shane addressed this

contracts/collections/sg721-base/src/contract.rs
yubrew: If current_royalty is `None`, there is no `RoyaltyInfo.updated_at` to compare

contracts/collections/sg721-base/src/contract.rs
yubrew: i think same rate limit for increasing / decreasing is fine.

contracts/collections/sg721-base/src/contract.rs
yubrew: need to migrate code ids for previous minters too

packages/sg721/src/lib.rs
yubrew: needs Option, fixed.

PR #578, opened 2023-06-09
Add Logic and Tests, Increase Royalties

yubrew:


contracts/collections/sg721-base/src/contract.rs
```diff
@@ -229,27 +229,57 @@ where
...
                     return Err(ContractError::RoyaltyShareIncreasedTooMuch {});
                 }
+
+                // if royalty share changed and current time is after last royalty update + min duration
+                // royalty_updated_at is always Some because it is set in instantiate
+                if (new_royalty_info_res.share != curr_royalty_info_res.share)
+                    && (curr_royalty_info_res
+                        .updated_at
+                        .plus_seconds(royalty_min_time_duration)
+                        > env.block.time)
```
shanev, line 269: Should this be `<`?

packages/sg2/src/msg.rs
```diff
@@ -30,6 +30,9 @@ pub struct UpdateMinterParamsMsg<T> {
     pub min_mint_price: Option<Coin>,
     pub mint_fee_bps: Option<u64>,
     pub max_trading_offset_secs: Option<u64>,
+    pub max_royalty_bps: Option<u64>,
+    pub max_royalty_increase_rate_bps: Option<u64>,
+    pub royalty_min_time_duration_secs: Option<u64>,
```
shanev, line 35: Instead of passing this in, I think its safe to assume 24 hours.

shanev (commented):
Mostly LGTM, just a few comments.

contracts/collections/sg721-base/src/contract.rs
yubrew: `if curr royalty updated_at + 24 hours > block time, throw error` I think that's right.

packages/sg2/src/msg.rs
yubrew: refactored to hardcode 24 hours

PR #585, opened 2023-06-28
Burn to Mint

humanalgorithm:
Closing in favor of https://github.com/public-awesome/launchpad/pull/610 due to rebase errors on this branch

This PR has several modules to it: 
1) Fungible and Nonfungible msg types in sg2: 
```rs
pub enum Token {
    Fungible(Coin),
    NonFungible(String),
}
```
2) The branching logic to check if receiving fungible or nonfungible token in contracts. All of the use of mint price through contracts and factories had to be changed to support this. Usually this is done implicitly in the .amount() checks. For example:
```rs
let config_mint_price = config.mint_price.clone().amount()?;
```
The amount() function has an implicit check: 
```rs
    pub fn amount(self) -> Result<Uint128, StdError> {
        let amount = match self {
            Token::Fungible(coin) => coin.amount,
            Token::NonFungible(_) => {
                return Err(StdError::generic_err("non-fungible tokens have no amount"))
            }
        };
        Ok(amount)
    }
}

```
3) Refactor of open edition minter tests: The open edition minter tests were in a bad state, so I refactored them to be more consistent. The files changed are the following: 

```rs
test-suite/src/common_setup/templates.rs
* in test-suite/src/open_edition_factory/
* in test-suite/src/open_edition_minter
* in test-suite/src/common_setup/setup_minter/open_edition_minter/
test-suite/src/common_setup/setup_minter/open_edition_minter.rs
```

4) Missing sudo tests: there was not sufficient tests for sudo update params so I added them in the following files: 

```rs
test-suite/src/base_factory/tests/sudo_tests.rs
test-suite/src/open_edition_factory/tests/sudo_tests.rs
test-suite/src/vending_factory/tests/sudo_tests.rs
```

5) Updates to minters: All three minters base-minter, vending-minter and open-edition minter had the following changes: 
- Add a  ReceiveNft(Cw721ReceiveMsg) path to execute
- Add a `execute_burn_to_mint` function to respond to the receive NFT message
- Updates to execute_mint functions to accomodate for not paying a mint fee when an NFT is received (collection contract will be calling itself)

6) `check_minter_caller` added to s721-base: We need this assertion check because, when an NFT is received, that contract will burn that NFT and then itself call a collection for a mint. The check minter caller, allows an sg721-base collection to call itself. 

7) burn_to_mint tests: burn_to_mint.rs was added to vending-minter, open-edition-minter and base-minter tests. A send nft is sent like so: 
```rs
    let send_nft = Cw721ExecuteMsg::SendNft {
        contract: minter_addr_2.to_string(),
        token_id: 1.to_string(),
        msg: to_binary("this is a test").unwrap(),
    };
    ```
    The message is sent to collection 1, with instructions to have collection 1 forward that NFT to minter 2. In this way, the functionality of burn to mint is tested end to end.

### CODE REVIEW

**Comment by MightOfOaks** [comment]
The creators will probably want to select which collection(s) the token(s) should be burned from in order for a token to be minted from their collection. I think we need to update the `MinterCreateMsg`s to optionally receive this information and check which collection the `Cw721ReceiveMsg` originates from in the `execute_burn_to_mint` functions.

Also not sure if we should update `mint_price` to be optional as well to allow for Burn-to-Mint-only collections.

**Comment by MightOfOaks** [comment]
Using this functionality with the 1/1 collections (base factory/minter) may not be ideal because the tokens are minted to the collection owner's wallet anyway. The current mint price is 5 STARS/token. Not sure if there is a use case for that.

**Comment by humanalgorithm** [comment]
> Using this functionality with the 1/1 collections (base factory/minter) may not be ideal because the tokens are minted to the collection owner's wallet anyway. The current mint price is 5 STARS/token. Not sure if there is a use case for that.

As discussed in Discord, we will leave the code in the base minter, not because it is likely to be used but because it will be easier to refactor in the future if the code is consistent across minters.

**Comment by humanalgorithm** [comment]
> The creators will probably want to select which collection(s) the token(s) should be burned from in order for a token to be minted from their collection. I think we need to update the `MinterCreateMsg`s to optionally receive this information and check which collection the `Cw721ReceiveMsg` originates from in the `execute_burn_to_mint` functions.
> 
> Also not sure if we should update `mint_price` to be optional as well to allow for Burn-to-Mint-only collections.

As discussed in Discord: 
- We will need an optional parameter that gives the list of valid collection addresses for burn to mint, rather than to just receive NFTs from random collections for payment
- The above should be added to the CreateMinterMsg as optional parameter
- Mint price will remain for now, with a future PR to add this. Reasoning being that it would make the PR too complicated to add this.

File: contracts/factories/base-factory/Cargo.toml [hunk 1]
```diff
@@ -35,5 +35,7 @@ schemars        = { workspace = true }
 serde           = { workspace = true }
 sg1             = { workspace = true }
 sg2             = { workspace = true }
+cw721-base = { workspace = true }
```
shanev, line 38:
```suggestion
cw721-base = { workspace = true, features = ["library"] }
```
Be sure to import as a library, not a full contract.

contracts/factories/base-factory/Cargo.toml
```diff
@@ -35,5 +35,7 @@ schemars        = { workspace = true }
...
+sg721-base      = { workspace = true}
```
shanev, line 39:
```suggestion
sg721-base      = { workspace = true, features = ["library"] } 
```

contracts/factories/vending-factory/src/contract.rs
```diff
@@ -82,15 +82,17 @@ pub fn execute_create_minter(
             got: msg.init_msg.per_address_limit,
         });
     }
-
+    //TODO BURNMINT
```
shanev, line 85: comment needed?

contracts/minters/base-minter/src/contract.rs
```diff
@@ -107,61 +111,145 @@ pub fn execute(
...
+    _execute_mint_sender(deps, env, info, token_uri)
+}
+
+fn _check_sender_is_collection_or_contract(
+    this_contract: Addr,
+    info: MessageInfo,
+    collection_info: CollectionInfoResponse,
+) -> Result<bool, ContractError> {
+    let sender_is_collection_or_contract =
+        vec![collection_info.creator, this_contract.to_string()].contains(&info.sender.to_string());
```
shanev, line 173: The contract itself can never be the sender. This is not a smart contract account or anything. So this can be removed and we can go back to just checking if the sender is the creator.

contracts/minters/open-edition-minter/src/contract.rs
```diff
@@ -599,23 +680,52 @@ pub fn execute_update_per_address_limit(
         .add_attribute("limit", per_address_limit.to_string()))
 }
 
+pub fn execute_burn_to_mint(
```
shanev, line 683: This code is duplicated. I would import it from another contract, or move it to packages (as a controller), and import that.

contracts/minters/vending-minter/src/contract.rs
```diff
@@ -243,9 +245,37 @@ pub fn execute(
             execute_update_discount_price(deps, env, info, price)
         }
         ExecuteMsg::RemoveDiscountPrice {} => execute_remove_discount_price(deps, info),
+        ExecuteMsg::ReceiveNft(msg) => execute_burn_to_mint(deps, env, info, msg),
     }
 }
 
+pub fn execute_burn_to_mint(
```
shanev, line 252: Duplicate code. Import from another contract or package.

contracts/minters/vending-minter/src/contract.rs
```diff
@@ -568,6 +611,62 @@ pub fn execute_mint_for(
     )
 }
 
+fn _pay_mint_if_not_contract(
+    this_contract: Addr,
```
shanev, line 615: Also seems like duplicate code that's in other contracts.

contracts/factories/vending-factory/src/contract.rs
humanalgorithm: Removed

contracts/factories/base-factory/Cargo.toml
humanalgorithm: updated

contracts/factories/base-factory/Cargo.toml
humanalgorithm: updated

contracts/minters/base-minter/src/contract.rs
humanalgorithm: I don't think this is correct.  
When run `fn execute_burn_to_mint` we first do a burn: 

```rs
    let cosmos_burn_msg = CosmosMsg::Wasm(WasmMsg::Execute {
        contract_addr: info.sender.to_string(),
        msg: to_binary(&burn_msg)?,
        funds: vec![],
    });
```
and then tack on a mint: 
```rs

    let cosmos_mint_msg = CosmosMsg::Wasm(WasmMsg::Execute {
        contract_addr: env.contract.address.to_string(),
        msg: to_binary(&execute_mint_msg)?,
        funds: vec![],
    });
    let res = res.add_message(cosmos_mint_msg);
    Ok(res)
   ```
   
   That mint message will show up as the minter executing a mint on itself and so you will end up in `pub fn execute_mint_sender` which is where we need to check if the mint message came from this minter and let it through.

File: contracts/minters/open-edition-minter/src/contract.rs (hunk 5 above)
Comment (humanalgorithm) [reply contracts/minters/open-edition-minter/src/contract.rs:683]: I'm not sure about this. The contract needs to call itself with a Mint Msg, which would have to match the MintMsg type expected for the contract (VendingMinter MintMsg for Vending Minter, OpenEditionMinter MintMsg for Open Edition Minter) etc. 

Or is there some way to put a generic as a parameter? 
This is the line in execute_burn_to_mint that would need a generic type in order to be packageable: 
```rs
    let execute_mint_msg = ExecuteMsg::Mint {
        token_uri: token_uri_msg.token_uri,
    };
    ```

File: contracts/minters/base-minter/src/contract.rs [hunk 8]
```diff
@@ -107,61 +111,146 @@ pub fn execute(
...
         &Sg721QueryMsg::CollectionInfo {},
     )?;
-    // allow only sg721 creator address to mint
-    if collection_info.creator != info.sender {
-        return Err(ContractError::Unauthorized(
-            "Sender is not sg721 creator".to_owned(),
-        ));
-    };
+
+    let _ = _check_sender_is_collection_or_contract(
```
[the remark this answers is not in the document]
tasiov: when would collection be sender?

contracts/minters/open-edition-minter/Cargo.toml
```diff
@@ -36,6 +36,7 @@ serde                   = { workspace = true }
 sg1                     = { workspace = true }
 sg2                     = { workspace = true }
 sg4                     = { workspace = true }
+cw721           = { workspace = true }
```
tasiov, line 39: formatting

tasiov (changes requested):

contracts/factories/open-edition-factory/src/contract.rs
```diff
@@ -76,10 +76,10 @@ pub fn execute_create_minter(
     if NATIVE_DENOM != msg.init_msg.mint_price.denom {
         return Err(ContractError::BaseError(BaseContractError::InvalidDenom {}));
     }
-
-    if params.min_mint_price.amount > msg.init_msg.mint_price.amount {
+    let min_mint_price = params.min_mint_price.amount()?;
```
yubrew, line 79: why change this?

contracts/minters/base-minter/src/contract.rs
humanalgorithm: This comment is unaddressable per Discord: 

```rs
"the minters do in fact call themselves when they send a mint message to reply which ends up back in the contract and I need to check if sender is self contract"
```

contracts/minters/open-edition-minter/src/contract.rs
humanalgorithm: Marking this as unaddressable comment per Discord: 

<blockquote>
I can't make this generic. for example in the base minter there is the helpers.rs that uses generics. 
</blockquote>

```rs
use crate::msg::ExecuteMsg;

  pub fn call<T: Into<ExecuteMsg>>(&self, msg: T) -> StdResult<CosmosMsg> 
  let msg = to_binary(&msg.into())?;
```

<blockquote>
Here base minter knows how to convert into an execute msg bc it is using the crate from the same package. We can't do that if burn to mint would be in another package, it would only be able to use its own msg/execute msg. 

Need to be able to convert into execute msg so that would be able to put it into a wasm msg something like this: 
</blockquote>

```rs
        Ok(WasmMsg::Execute {
            contract_addr: self.addr().into(),
            msg,
            funds: vec![funds],
        }
 ```

<blockquote>
Then if the mint msg is generic as T it doesn't know how to convert it to binary
</blockquote>

```rs
let execute_msg = to_binary(&execute_mint_msg)?;
the trait bound `T: Serialize` is not satisfied
the trait `Serialize` is not implemented for `T`
```
<blockquote>
so even if you try to pass in the execute msg already formed into another package it won't know what to do with it
</blockquote>

contracts/minters/vending-minter/src/contract.rs
humanalgorithm: Resolved per this can not be put in another package.

contracts/minters/open-edition-minter/src/contract.rs
```diff
@@ -264,12 +266,93 @@ pub fn execute_mint_to(
...
+    deps: &DepsMut,
+    this_contract: Addr,
+    info: MessageInfo,
+    mint_price_with_discounts: Coin,
+    mint_fee: Decimal,
+    factory_params: MinterParams<open_edition_factory::state::ParamsExtension>,
+) -> Result<(Response, Uint128), ContractError> {
+    let mut res = Response::new();
+    match info.clone().sender == this_contract {
+        true => Ok((res, Uint128::new(0))),
```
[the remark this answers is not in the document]
humanalgorithm: It has to be, because when you send an NFT as payment, you are not also sending token. So what would you be fair burning? You can't fair burn an NFT.

contracts/factories/base-factory/Cargo.toml
```diff
@@ -35,5 +35,7 @@ schemars        = { workspace = true }
 serde           = { workspace = true }
 sg1             = { workspace = true }
 sg2             = { workspace = true }
+cw721-base = { workspace = true, features = ["library"] }
```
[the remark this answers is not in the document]
humanalgorithm: Fixed

contracts/collections/sg721-base/src/contract.rs
```diff
@@ -313,7 +315,11 @@ where
         info: MessageInfo,
         nft_data: NftParams<T>,
     ) -> Result<Response, ContractError> {
-        assert_minter_owner(deps.storage, &info.sender)?;
+        let is_minter_caller =
```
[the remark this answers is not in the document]
humanalgorithm: As discussed in Discord, it seems I don't need to use the second method of looking for the owner through self.parent.query, therefore I can drop the redundant function. Everything is going through `assert_minter_owner` now.

contracts/collections/sg721-base/src/contract.rs
```diff
@@ -322,7 +328,7 @@ where
                 extension,
             } => (token_id, owner, token_uri, extension),
         };
-
+        deps.api.addr_validate(&owner)?;
```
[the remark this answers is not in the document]
humanalgorithm: True, don't need this. Dropping.

contracts/collections/sg721-base/src/contract.rs
```diff
@@ -350,6 +355,18 @@ where
...
     }
 
+    pub fn check_minter_caller(
+        &self,
+        deps: Deps,
+        env: Env,
+        sender: Addr,
+    ) -> Result<bool, ContractError> {
+        let msg = msg::QueryMsg::Minter {};
+        let minter: MinterResponse = from_binary(&self.parent.query(deps, env, msg.into())?)?;
```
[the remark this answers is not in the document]
humanalgorithm: Yepp. Dropping this function in favor of `assert_minter_owner`

contracts/minters/base-minter/src/contract.rs
```diff
@@ -107,61 +111,146 @@ pub fn execute(
...
+
 pub fn execute_mint_sender(
     deps: DepsMut,
+    env: Env,
+    info: MessageInfo,
+    token_uri: String,
+) -> Result<Response, ContractError> {
+    let config = CONFIG.load(deps.storage)?;
+
+    if matches!(config.mint_price, Token::NonFungible(_)) {
```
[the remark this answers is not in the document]
humanalgorithm: Nope, mint_price is a token, token looks like this: 
```rs
pub enum Token {
    Fungible(Coin),
    NonFungible(String),
}
```
So if you get a NonFungible token it will fail, but Fungible token won't match the pattern and therefore this check will pass.

contracts/minters/base-minter/src/contract.rs
```diff
@@ -107,61 +111,146 @@ pub fn execute(
...
+
+    _execute_mint_sender(deps, env, info, token_uri)
+}
+
+fn _check_sender_is_collection_or_contract(
+    this_contract: Addr,
+    info: MessageInfo,
+    collection_info: CollectionInfoResponse,
+) -> Result<bool, ContractError> {
+    println!("sender is {:?}", info.sender);
```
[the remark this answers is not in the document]
humanalgorithm: Removed.

contracts/minters/base-minter/src/contract.rs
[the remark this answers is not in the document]
humanalgorithm: > no need to return value if you are not going to use

Removed.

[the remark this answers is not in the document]
humanalgorithm: > when would collection be sender?

Ahh you know what, this is not really named properly. Its more like "check_sender_is_creator_or_contract". Creator as the owner, and then contract when it calls itself with the burn and mint messages from execute burn to mint.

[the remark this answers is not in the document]
humanalgorithm: Renamed to `check_sender_is_creator_or_contract`

contracts/minters/base-minter/src/contract.rs
```diff
@@ -107,61 +111,146 @@ pub fn execute(
...
+        env.contract.address,
+        info.clone(),
+        config.mint_price,
+        factory_params,
+    )?;
     // Create mint msgs
     let mint_msg = Sg721ExecuteMsg::<Extension, Empty>::Mint {
         token_id: increment_token_index(deps.storage)?.to_string(),
-        owner: info.sender.to_string(),
+        owner: collection_info.creator,
```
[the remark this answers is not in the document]
humanalgorithm: > the NFT owner would be the collection creator?

So now that the contract can call itself through the execute_burn_to_mint messages, we can't just assume the the sender is always creator, and therefore info.sender won't cut it.

Actually looking at the name of this function `execute_mint_sender` I'm not entirely sure what is the intention of this function. It sounds like a mint is being performed on behalf of the sender? But here I am just putting back the collection creator because that's all I have access to to figure out who should be the owner. Unless I do a query for owner here? Thoughts?

contracts/minters/open-edition-minter/Cargo.toml
humanalgorithm: Updated

contracts/minters/base-minter/src/contract.rs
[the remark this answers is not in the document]
humanalgorithm: This redundant check was removed per Discord convo

test-suite/src/open_edition_minter/tests/ibc_asset_mint.rs
```diff
@@ -57,8 +73,7 @@ fn check_custom_create_minter_denom() {
         .ok();
 
     setup_block_time(&mut router, GENESIS_MINT_START_TIME + 100, None);
-
-    // Mint succeeds
+    //     // Mint succeeds
```
yubrew, line 76:
```suggestion
    // Mint succeeds
```

PR #595, opened 2023-07-20
Royalty update fix

tasiov:


jhernandezb (changes requested):
- add constants for harcoded values
- allow decreasing royalties without percent check

contracts/collections/sg721-base/src/contract.rs
```diff
@@ -221,35 +225,45 @@ where
...
+                        )));
+                    }
+                    if new_royalty_info.share > Decimal::percent(MAX_ROYALTY_SHARE_PCT) {
+                        return Err(ContractError::InvalidRoyalties(
+                            format!("Share cannot be greater than {}%", MAX_ROYALTY_SHARE_PCT)
+                                .to_string(),
+                        ));
+                    }
+                } else {
                 }
```
jhernandezb, line 260: let's remove this else

contracts/collections/sg721-base/src/contract.rs
```diff
@@ -366,40 +380,50 @@ where
...
     }
 }
 
 pub fn share_validate(share: Decimal) -> Result<Decimal, ContractError> {
     if share > Decimal::one() {
-        return Err(ContractError::InvalidRoyalties {});
+        return Err(ContractError::InvalidRoyalties(
+            "Share cannot be greater than 100%".to_string(),
+        ));
     }
```
jhernandezb, line 427: Decimal type allows only positive numbers right ?

jhernandezb (approved):

contracts/collections/sg721-base/src/contract.rs
tasiov: yes Decimal is Uint under the hood

Screenshot 2023-07-31 at 11 07 18 AM

PR #596, opened 2023-07-24
SG-997 Add IBC denoms to open edition

yubrew:
ibc denoms for open edition minter (`frenz`)

I ran e2e tests locally and they passed

### DIFF
--- a/contracts/factories/open-edition-factory/src/contract.rs
+++ b/contracts/factories/open-edition-factory/src/contract.rs
@@ -92,7 +92,7 @@ pub fn execute_create_minter(
     );
 
     ensure!(
-        params.min_mint_price.amount >= msg.init_msg.mint_price.amount,
+        params.min_mint_price.amount <= msg.init_msg.mint_price.amount,
         ContractError::InsufficientMintPrice {
             expected: params.min_mint_price.amount.u128(),
             got: msg.init_msg.mint_price.amount.into(),
--- a/contracts/factories/open-edition-factory/src/contract.rs
+++ b/contracts/factories/open-edition-factory/src/contract.rs
@@ -33,17 +33,19 @@ pub fn instantiate(
     msg: InstantiateMsg,
 ) -> Result<Response, ContractError> {
     set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+    let params = msg.params;
 
-    // airdrop fee and mint fee should be same denom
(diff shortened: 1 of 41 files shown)


contracts/factories/open-edition-factory/src/contract.rs
```diff
@@ -32,6 +34,15 @@ pub fn instantiate(
 ) -> Result<Response, ContractError> {
     set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
```
shanev, line 36: This could be written a lot cleaner. 

1. Please start using `ensure!()` for these kinds of checks.
2. Having 3+ deep calls is a code smell, so can make cleaner with `let params = msg.params`.
3. `ContractError` is imported, so no reason to duplicate it in error calls.

shanev (changes requested):
LGTM besides some minor refactoring.

I wanted to start using Typescript for e2e tests but since this already has Rust-based tests, it's fine that it's continued. I understand that its beyond the scope of this PR to do that.

contracts/factories/open-edition-factory/src/contract.rs
yubrew: fixed

shanev (approved):

shanev:
@yubrew We can update to use the Fair Burn contract now since its on mainnet yeah?

contracts/minters/open-edition-minter/src/contract.rs
```diff
@@ -313,16 +313,36 @@ fn _execute_mint(
...
+                        .addr_validate(&factory_params.extension.dev_fee_address)?,
+                ),
+                &mut res,
+            )?;
+            // let msg =
+            //     create_fund_community_pool_msg(coins(network_fee.u128(), mint_price.clone().denom));
+            // res = res.add_message(msg);
+        }
+    } else {
+        checked_fair_burn(
```
tasiov, line 336: i guess this branch should also have the network fee zero check?

yubrew: yea you're right, should check 0fee

yubrew:
> @yubrew We can update to use the Fair Burn contract now since its on mainnet yeah?

config.fee_percent is 50%, this contract needs a different percent.
https://github.com/public-awesome/core/blob/main/contracts/fair-burn/src/helpers.rs#L9

PR #597, opened 2023-07-25
Update sg721-updatable migration logic

MightOfOaks:


yubrew (approved):
nice work. fixes bug where sg721-base wants to migrate to sg721-updatable with same contract version.

PR #605, opened 2023-08-19
Package error for burn to mint [CODE-SHARE]

humanalgorithm:


shanev:
Closing because we're going to do burn-to-mint using pre-mint hooks instead.

PR #609, opened 2023-08-31
SG-1034 send ibc denom fairburns to foundation addr

yubrew:


tasiov (approved):
Hardcoded address doesn't feel great, but lgtm otherwise

shanev:
Let's use a new multisig address for better accounting. Also I see this as temporary until we move to using the Fair Burn contract everywhere.

shanev (approved):
Will wait to merge until the community sentiment for this is positive in Commonwealth.

PR #613, opened 2023-09-06
Added End Time and Num Tokens Options within the Open Edition Minter/Factory

racoon-supply-rac:
Following some requests, we added a bit more flexibility to the OE Minter/Factory.

You can now have the Mint End Time as optional and you also have a Max Number of Tokens to be minted which is also optional.

There are no real breaking change except that the End Time is now an `Option` instead of a Timestamp when stored in the states.

We're open to all comments and feedback to make some change to make it easier to integrate on Stargaze.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -278,7 +278,7 @@ dependencies = [
 
 [[package]]
 name = "base-factory"
-version = "3.2.0"
+version = "3.2.9"
 dependencies = [
  "cosmwasm-schema",
  "cosmwasm-std",
@@ -295,7 +295,7 @@ dependencies = [
 
 [[package]]
 name = "base-minter"
-version = "3.2.0"
+version = "3.2.9"
 dependencies = [
  "base-factory",
  "cosmwasm-schema",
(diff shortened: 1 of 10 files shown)


humanalgorithm:
@racoon-supply-rac humanalgorithm here. I'm the one who originally wrote the template structure for tests. To be consistent with the rest of the test-suite we should have it match that.

I had an open edition minter tests refactor that got delayed, it's now here:
https://github.com/public-awesome/launchpad/pull/614

I can work with you to get the tests here until the format of the above PR. You can message me on Discord @humanalgorithm same name. 

In the future we might do another round of refactor, but for now should keep it consistent.

racoon-supply-rac:
> @racoon-supply-rac humanalgorithm here. I'm the one who originally wrote the template structure for tests. To be consistent with the rest of the test-suite we should have it match that.
> 
> I had an open edition minter tests refactor that got delayed, it's now here: #614
> 
> I can work with you to get the tests here until the format of the above PR. You can message me on Discord @humanalgorithm same name.
> 
> In the future we might do another round of refactor, but for now should keep it consistent.

Great! Thanks for letting us know. We'll have a look soon at the new test structure and poke you on Discord if needed

PR #627, opened 2023-11-27
token URI scheme as init arg

dgabriele:


shanev (approved):
LGTM, makes Launchpad more flexible.

shanev (commented):
Actually we can update this to not even store the URL scheme in the factory. We can remove the URL check altogether to give creators more freedom. Frontend clients can display the URL to inform the user about the type of storage.

jhernandezb:
closing in favor of #643

PR #629, opened 2023-12-07
Open Edition minter update & tests

MightOfOaks:
Allows the user to set a limit for the number of tokens instead of a limit for minting end time.


contracts/minters/open-edition-minter/src/contract.rs
```diff
@@ -193,10 +200,20 @@ pub fn execute_purge(
     info: MessageInfo,
 ) -> Result<Response, ContractError> {
     nonpayable(&info)?;
-    // Check if mint has ended
+    // check if sold out (optional)
+    let mintable_num_tokens = MINTABLE_NUM_TOKENS.may_load(deps.storage)?;
```
tasiov, line 204: so what if it's ended but not sold out?

contracts/minters/open-edition-minter/src/state.rs
```diff
@@ -11,8 +11,9 @@ pub struct ConfigExtension {
     pub payment_address: Option<Addr>,
     pub nft_data: NftData,
     pub start_time: Timestamp,
-    pub end_time: Timestamp,
+    pub end_time: Option<Timestamp>,
```
tasiov, line 14: curious if this will cause errors on read

tasiov (approved):

contracts/minters/open-edition-minter/src/contract.rs
MightOfOaks: `mintable_num_tokens` is only Some when there is no end time defined (as per instantiation), so the minting cannot be time limited in this case.

contracts/minters/open-edition-minter/src/state.rs
MightOfOaks: The contract will be used for new open edition collections onwards without any migrations. It could cause indexing errors though if not accounted for.

jhernandezb (approved):

PR #630, opened 2023-12-11
  BE-239 - Include mint price denom among mint event attributes for minter-wl-flex

MightOfOaks:


contracts/minters/vending-minter-wl-flex/src/contract.rs
```diff
@@ -730,9 +730,15 @@ fn _execute_mint(
         .add_attribute("sender", info.sender)
         .add_attribute("recipient", recipient_addr)
         .add_attribute("token_id", mintable_token_mapping.token_id.to_string())
-        .add_attribute("network_fee", network_fee)
-        .add_attribute("mint_price", mint_price.amount)
-        .add_attribute("seller_amount", seller_amount))
+        .add_attribute(
+            "network_fee",
+            coin(network_fee.u128(), mint_price.clone().denom).to_string(),
```
jhernandezb, line 735: I don't think we need to change network fee if it is a coin already the to_string() implementation should be including amount and denom

contracts/minters/vending-minter-wl-flex/src/contract.rs
```diff
@@ -730,9 +730,15 @@ fn _execute_mint(
...
+        )
+        .add_attribute("mint_price", mint_price.to_string())
+        .add_attribute(
+            "seller_amount",
+            coin(seller_amount.u128(), mint_price.denom).to_string(),
```
jhernandezb, line 740: same here no need to do `coin(..)` again

contracts/minters/vending-minter-wl-flex/src/contract.rs
```diff
@@ -730,9 +730,15 @@ fn _execute_mint(
...
         .add_attribute("recipient", recipient_addr)
         .add_attribute("token_id", mintable_token_mapping.token_id.to_string())
-        .add_attribute("network_fee", network_fee)
-        .add_attribute("mint_price", mint_price.amount)
-        .add_attribute("seller_amount", seller_amount))
+        .add_attribute(
+            "network_fee",
+            coin(network_fee.u128(), mint_price.clone().denom).to_string(),
+        )
+        .add_attribute("mint_price", mint_price.to_string())
```
jhernandezb, line 737: this looks good although not sure if to_string is necessary

jhernandezb (changes requested):

contracts/minters/vending-minter-wl-flex/src/contract.rs
MightOfOaks: Receiving a `the trait From<Coin> is not implemented for std::string::String` error otherwise.

contracts/minters/vending-minter-wl-flex/src/contract.rs
MightOfOaks: `network_fee` is only a Uint128 here I am afraid.

contracts/minters/vending-minter-wl-flex/src/contract.rs
MightOfOaks: Same for `seller_amount`, it's a Uint128.

tasiov (approved):

jhernandezb (approved):

PR #631, opened 2023-12-12
Royalty increase update for v1 collections

MightOfOaks:


contracts/sg721-updatable/src/contract.rs
```diff
@@ -295,19 +316,8 @@ pub fn migrate(deps: DepsMut, _env: Env, _msg: Empty) -> Result<Response, Contra
...
-            payment_address: royalty_info.payment_address,
-            share: royalty_info.share,
-            // setting updated_at to 1 day ago to make updating royalties possible right after a migration
-            updated_at: Some(_env.block.time.minus_seconds(86400)),
-        };
-        collection_info.royalty_info = Some(new_royalty_info);
-        COLLECTION_INFO.save(deps.storage, &collection_info)?;
-    }
+    // setting updated_at to 1 day ago to make updating royalties possible right after a migration
+    ROYALTY_UPDATED_AT.save(deps.storage, &_env.block.time.minus_seconds(86400))?;
```
jhernandezb, line 320: rename the variable to `env` usually `_` is for unused variables but when you switch to use it ideally you rename the param in the method.

jhernandezb (changes requested):

contracts/sg721-updatable/src/state.rs
```diff
@@ -16,8 +16,8 @@ pub struct CollectionInfo<T> {
 pub struct RoyaltyInfo {
     pub payment_address: Addr,
     pub share: Decimal,
-    pub updated_at: Option<Timestamp>,
```
tasiov, line 19: just want to check with @MightOfOaks to make sure this won't break on read

contracts/sg721-updatable/src/contract.rs
```diff
@@ -27,6 +29,7 @@ const CONTRACT_VERSION: &str = env!("CARGO_PKG_VERSION");
 
 const CREATION_FEE: u128 = 1_000_000_000;
 const MAX_ROYALTY_SHARE_BPS: u64 = 1_000;
+const MAX_ROYALTY_SHARE_DELTA_BPS: u64 = 200;
```
tasiov, line 32: this will be a PITA to update, but this may be how we must do it for now

contracts/sg721-updatable/src/contract.rs
```diff
@@ -147,40 +150,59 @@ pub fn execute_update_royalty_info(
...
-
-        if share > curr_royalty_info.share && update_too_soon {
-            return Err(ContractError::RoyaltyUpdateTooSoon {});
+        ensure!(
+            end.is_expired(&env.block),
+            ContractError::RoyaltyUpdateTooSoon {}
+        );
+        // Make sure the share is not increased more than MAX_ROYALTY_SHARE_DELTA_BPS at a time
+        if share > old_royalty_info.share {
+            let share_delta = share - old_royalty_info.share;
```
tasiov, line 174: what if `share < old_royalty_info.share`

tasiov (approved):

contracts/sg721-updatable/src/state.rs
MightOfOaks: Earlier the code was in sync with a v2 update that we abandoned and was never deployed on chain. So, the v1 contracts that are currently on chain do not have an `updated_at` attribute in their CollectionInfo.RoyaltyInfo.

The PR basically reverts all the previous changes regarding royalty updates and keeps the update_time separately similar to how it's done for the v2 contracts.

contracts/sg721-updatable/src/contract.rs
MightOfOaks: The reasoning was that the upwards updates are the risky ones that may cause traders to lose money if the change is too steep.

jhernandezb (approved):

PR #633, opened 2024-01-15
Allow updating collection creator

taitruong:


contracts/collections/sg721-base/src/contract.rs
```diff
@@ -207,6 +207,10 @@ where
             return Err(ContractError::Unauthorized {});
         }
 
+        if let Some(new_creator) = collection_msg.creator {
```
jhernandezb, line 210: would be nice to add addr_validate before storing it.

jhernandezb (changes requested):

contracts/collections/sg721-base/src/contract.rs
taitruong: LOL. You r right, ofc :).

PR #652, opened 2024-01-30
dYdX Airdrop updates

MightOfOaks:


contracts/dydx-airdrop/src/build_messages.rs
```diff
@@ -0,0 +1,140 @@
...
+    let bank_msg = SubMsg::new(BankMsg::Send {
+        to_address: info.sender.to_string(),
+        amount: coins(airdrop_amount, NATIVE_DENOM),
+    });
+    res = res.add_submessage(bank_msg);
+
+    Ok(res)
+}
+
+pub fn dust_and_whitelist_add(
```
humanalgorithm, line 69: I think this can just be whitelist_add and call the dust( ) function from the calling function

contracts/dydx-airdrop/src/validation.rs
```diff
@@ -0,0 +1,205 @@
...
+        .unwrap_or(false);
+    if registered {
+        return Err(ContractError::AlreadyRegistered {
+            address: eth_address.to_string(),
+        });
+    }
+    Ok(())
+}
+
+pub fn check_previous_claim(deps: &DepsMut, eth_address: &str) -> Result<(), ContractError> {
```
humanalgorithm, line 197: Shouldn't a "previous claim" just be the address to mint count matches the max airdrop amount or do we need to track something different?

test-suite/src/dydx_airdrop/tests/test_claim.rs
```diff
@@ -37,6 +37,7 @@ fn query_minter_as_expected(app: &mut StargazeApp, airdrop_contract: Addr, minte
 fn test_instantiate() {
     let mut app = custom_mock_app();
     let minter_address = Addr::unchecked("contract1");
+    let name_discount_wl_address = Addr::unchecked(MOCK_NAME_DISCOUNT_WL_ADDR_STR);
```
humanalgorithm, line 40: I think we need some new unit tests that test the additional functionality, such as can only do the claim after name is registered correct?

contracts/dydx-airdrop/src/validation.rs
MightOfOaks: Only the reward is being claimed currently. Not tracking a mint count. The user is expected to mint the token outside of the airdrop contract through the regular minting process.

test-suite/src/dydx_airdrop/tests/test_claim.rs
MightOfOaks: Correct. Updating them currently.

PR #660, opened 2024-02-22
Add whitelist-immutable-flex

MightOfOaks:


contracts/whitelists/whitelist-immutable-flex/src/helpers.rs
```diff
@@ -0,0 +1,45 @@
...
+
+    pub fn includes(&self, querier: &QuerierWrapper, address: String) -> StdResult<bool> {
+        let includes: bool = querier.query(&QueryRequest::Wasm(WasmQuery::Smart {
+            contract_addr: self.addr().into(),
+            msg: to_json_binary(&QueryMsg::HasMember { address })?,
+        }))?;
+        Ok(includes)
+    }
+
+    pub fn address_count(&self, querier: &QuerierWrapper) -> StdResult<u64> {
```
humanalgorithm, line 29: Is there address count or is this mint count for address? I think the second name makes more sense?

contracts/whitelists/whitelist-immutable-flex/src/state.rs
```diff
@@ -0,0 +1,21 @@
...
+use cosmwasm_std::{Addr, Decimal, Uint128};
+use cw_storage_plus::{Item, Map};
+
+#[cw_serde]
+pub struct Config {
+    pub admin: Addr,
+    pub mint_discount_bps: Option<u64>,
+}
+
+impl Config {
```
humanalgorithm, line 11: Why do we need a whitelist wide discount bps?

contracts/whitelists/whitelist-immutable-flex/src/helpers.rs
MightOfOaks: There is a separate Member query that returns the mint count for each address. AddressCount is legacy and returns the total number of members.

contracts/whitelists/whitelist-immutable-flex/src/state.rs
MightOfOaks: It's an optional parameter and we can remove it, but my intention was not to take away from the original contract in case it's used again for its initial purpose but with a `mint_count` for each address.

PR #668, opened 2024-03-19
Migration from sg721 to cw721 `CollectionInfo`

taitruong:
# Migration from sg721 to cw721 `CollectionInfo`

All changes in this PR is based on new `cw721` package. Check for PR in `cw-nfts` repo [here](https://github.com/CosmWasm/cw-nfts/pull/156).

There are these naming conventions for metadata in cw721: collection info, nft info, and extensions:
- collection info: collection infos (aka contract info) is required for a contract (containing name and symbol)
- collection extension: optional extension (containing description, image, etc.)
- nft extension: is optional metadata for an nft

Here `sg721-base` is using new `cw721-base` with collection info. `CollectionInfo` (sg721) is migrated to `CollectionExtension` in cw721. `sg721-base` does not longer need to hold dedicated collection info store (legacy). As a backup legacy is kept - and may be removed in later versions.

As a result code footprint of `sg721-base` is a lot smaller :).

# Backward Compatibility

For backward compatibility and not breaking 3rd parties (like frontend) legacy msgs, structs, etc are kept - but marked as deprecated. Like this:

```rs
#[cw_serde]
#[deprecated = "Please use `CollectionInfo<CollectionExtensionResponse<RoyaltyInfo>>` instead"]
pub struct CollectionInfoResponse {
...
}

pub enum QueryMsg<
     // Return type of NFT metadata defined in `NftInfo` and `AllNftInfo`.
    TNftExtension,
    // Return type of collection extension defined in `GetCollectionInfo`.
    TCollectionExtension,
    // Custom query msg for custom contract logic. Default implementation returns an empty binary.
    TExtensionQueryMsg,
> {
    #[allow(deprecated)]
    #[returns(CollectionInfoResponse)]
     #[deprecated = "Please use GetCollectionInfo instead"]
    CollectionInfo {},
...
}

pub enum ExecuteMsg<pub enum ExecuteMsg<
    // NftInfo extension msg for onchain metadata.
    TNftExtensionMsg,
    // CollectionInfo extension msg for onchain collection attributes.
    TCollectionExtensionMsg,
    // Custom extension msg for custom contract logic. Default implementation is a no-op.
    TExtensionMsg,
> {
    // ---- sg721 specific msgs ----
    /// Update specific collection info fields
    #[deprecated = "Please use Cw721UpdateCollectionInfo instead"]
    UpdateCollectionInfo {
        #[allow(deprecated)]
        collection_info: UpdateCollectionInfoMsg<RoyaltyInfoResponse>,
    },
...
}

#[cw_serde]
#[deprecated = "Please use `UpdateCollectionInfo<DefaultOptionalCollectionExtensionMsg>` instead"]
pub struct UpdateCollectionInfoMsg<T> {
...
}
```

There are also some helper code like:

```rs
#[allow(deprecated)]
impl From<InstantiateMsg> for Cw721InstantiateMsg<DefaultOptionalCollectionExtensionMsg> {
...
}

impl<TNftExtensionMsg, TCollectionExtensionMsg>
    From<ExecuteMsg<TNftExtensionMsg, TCollectionExtensionMsg>> for Sg721ExecuteMsg
...

impl From<QueryMsg>
    for Sg721QueryMsg<DefaultOptionalNftExtension, DefaultOptionalCollectionExtension, Empty>
...

impl From<CollectionInfo<RoyaltyInfoResponse>> for DefaultOptionalCollectionExtensionMsg {
...
}

impl From<UpdateCollectionInfoMsg<RoyaltyInfoResponse>>
    for CollectionExtensionMsg<RoyaltyInfoResponse>
...

impl From<InstantiateMsg> for Cw721InstantiateMsg<DefaultOptionalCollectionExtensionMsg> {
...
}

```

# sg721 derivatives

`sg721-metadata-onchain` has been removed. new `cw721-base` supports it by default - hence `sg721-base` as well.

TBD: `sg721-updatable` could be removed, since `cw721-base` also supports this (`UpdateNftInfo`). `sg721-updatable` only updates token uri, while `cw721-base` also allows updating onchain metadata:

```rs
pub enum Cw721ExecuteMsg<
    // Message passed for updating metadata.
    TNftMetadataExtensionMsg,
    // Message passed for updating collection metadata extension.
    TCollectionMetadataExtensionMsg,
> {
...
/// The creator is the only one eligible to update NFT's token uri and onchain metadata (`NftInfo.extension`).
    /// NOTE: approvals and owner are not affected by this call, since they belong to the NFT owner.
    UpdateNftInfo {
        token_id: String,
        token_uri: Option<String>,
        extension: TNftMetadataExtensionMsg,
    },
...
}
```

# new `StateFactory`

As recently discussed there was a bug in royalty check:
- during instantiation, royalty can be set to max 100%
- during execution (update), royalty can be set to max 10%

A reason how this bug wasnt obivious is that royalty validation happend in 2 different places.

In `cw721` a new `StateFactory` has been introduced. Check PR comment [here](https://github.com/CosmWasm/cw-nfts/pull/156#discussion_r1532943024).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1367,7 +1367,7 @@ dependencies = [
 [[package]]
 name = "cw721"
 version = "0.19.0"
-source = "git+https://github.com/arkprotocol/cw-nfts.git?branch=collection-info#dea2b0175f77a2dabd25ad109244bc215f0740df"
+source = "git+https://github.com/arkprotocol/cw-nfts.git?branch=collection-info#1d2b176a5d310cb2031866f54a6dfe0348806f48"
 dependencies = [
  "cosmwasm-schema",
  "cosmwasm-std",
@@ -1422,7 +1422,7 @@ dependencies = [
 [[package]]
 name = "cw721-base"
 version = "0.19.0"
-source = "git+https://github.com/arkprotocol/cw-nfts.git?branch=collection-info#dea2b0175f77a2dabd25ad109244bc215f0740df"
+source = "git+https://github.com/arkprotocol/cw-nfts.git?branch=collection-info#1d2b176a5d310cb2031866f54a6dfe0348806f48"
 dependencies = [
  "cosmwasm-std",
  "cw-ownable 0.6.0",


contracts/collections/sg721-base/src/contract.rs
```diff
@@ -1,275 +1,172 @@
...
-        if let Some(Some(new_royalty_info_response)) = collection_msg.royalty_info {
+        if let Some(Some(_)) = collection_msg.royalty_info {
             let last_royalty_update = self.royalty_updated_at.load(deps.storage)?;
             if last_royalty_update.plus_seconds(24 * 60 * 60) > env.block.time {
                 return Err(ContractError::InvalidRoyalties(
                     "Royalties can only be updated once per day".to_string(),
                 ));
             }
 
-            let new_royalty_info = RoyaltyInfo {
```
humanalgorithm, line 243: How come we can delete all this royalty calculation?

contracts/collections/sg721-base/src/msg.rs
```diff
@@ -1,128 +1,178 @@
...
+
+    /// Returns `CollectionInfoAndExtensionResponse`
+    #[returns(CollectionInfoAndExtensionResponse<TCollectionExtension>)]
+    GetCollectionInfoAndExtension {},
+
+    /// returns `AllInfoResponse` which contains contract, collection and nft details
+    #[returns(AllInfoResponse)]
+    GetAllInfo {},
+
+    /// Returns `CollectionExtensionAttributes`
```
humanalgorithm, line 93: Can we remove these comments when it just says the same thing as the code/doesn't add any information?

packages/sg721/src/lib.rs
```diff
@@ -1,11 +1,50 @@
...
+    // NftInfo extension msg for onchain metadata.
+    TNftExtensionMsg,
+    // CollectionInfo extension msg for onchain collection attributes.
+    TCollectionExtensionMsg,
+    // Custom extension msg for custom contract logic. Default implementation is a no-op.
+    TExtensionMsg,
+> {
+    // ---- sg721 specific msgs ----
+    /// Update specific collection info fields
+    #[deprecated = "Please use Cw721UpdateCollectionInfo instead"]
```
humanalgorithm, line 27: A lot of allow deprecated everywhere. What would be the condition in which we can remove these tags? Will it ever meet that condition?

packages/sg721/src/lib.rs
```diff
@@ -51,27 +91,129 @@ pub enum ExecuteMsg<T, E> {
...
+    WithdrawFunds {
+        amount: Coin,
     },
-    /// Called by the minter to update trading start time
-    UpdateStartTradingTime(Option<Timestamp>),
-    // Freeze collection info from further updates
-    FreezeCollectionInfo,
+}
+
+impl<TNftExtensionMsg, TCollectionExtensionMsg, TExtensionMsg>
```
humanalgorithm, line 128: Is lib.rs the right place for this?

contracts/collections/sg721-base/src/contract.rs
```diff
@@ -1,275 +1,172 @@
...
+        TExtensionMsg,
+        TExtensionQueryMsg,
+        TCustomResponseMsg,
+    >
+    Sg721Contract<
+        'a,
+        // NftInfo extension (onchain metadata).
+        TNftExtension,
+        // NftInfo extension msg for onchain metadata.
+        TNftExtensionMsg,
```
humanalgorithm, line 39: RE naming conventions. I'm just curious if we need to name variables this way. The comment says NFTInfo extension msg for onchain metadata. Would ie be wrong to name it NFTInfoOnchainMetadataExtensionMsg or OnchainMetadataExtensionMsg something like that? I see this pattern in a lot of variables where the description is different than the name. Or do we need these names for consistency?

contracts/collections/sg721-updatable/src/contract.rs
```diff
@@ -7,19 +7,33 @@ use cosmwasm_std::{Deps, StdResult};
...
 
-use cw721_base::Extension;
 use cw_utils::nonpayable;
 use sg1::checked_fair_burn;
 use sg721_base::ContractError::Unauthorized;
 use sg721_base::Sg721Contract;
-pub type Sg721UpdatableContract<'a> = Sg721Contract<'a, Extension>;
+pub type Sg721UpdatableContract<'a> = Sg721Contract<
+    'a,
+    DefaultOptionalNftExtension,
```
humanalgorithm, line 29: What is meant by "default" ?

taitruong: It is defined in `cw721` package: `pub type DefaultOptionalNftExtension = Option<NftExtension>;`

So it is optional onchain metadata.

packages/sg721/src/lib.rs
taitruong: As noted in PR description:

```md
...
# Backward Compatibility
For backward compatibility and not breaking 3rd parties (like frontend) legacy msgs, structs, etc are kept - but marked as deprecated.
...
```

So this PR can be deployed without breaking frontends. Once frontends (e.g. launchpad, marketplace) are migrated, then all deprecated types can be removed.

packages/sg721/src/lib.rs
taitruong: 100%, before PR changes `sg721` isnt suppose to hold much specific data or logic - since everything is in `cw721`. so lib.rs in here just binds cw721 and adds only few additional sg specific data.

contracts/collections/sg721-base/src/contract.rs
taitruong: That is imo the cool part about it, it is moved to cw721. RoyalInfoMsg implements StateFactory where in `create()` it also does `validate()`.

Check in other PR my comment here: https://github.com/CosmWasm/cw-nfts/pull/156/files#r1541437296

contracts/collections/sg721-base/src/contract.rs
```diff
@@ -1,275 +1,172 @@
...
-        self.collection_info.save(deps.storage, &collection)?;
+        let collection_extension: CollectionExtensionMsg<RoyaltyInfoResponse> =
+            collection_msg.into();
+        let msg = CollectionInfoMsg {
+            name: None,
+            symbol: None,
+            extension: Some(collection_extension),
+        };
+        self.parent
+            .update_collection_info(deps, Some(&info), Some(&env), msg)?;
```
taitruong, line 169: Here logic is delegated to parent (cw721 package), where collection info and royalty msg is passed and cw721 just do this call:

`let collection_info = msg.create(deps.as_ref().into(), env, info, Some(&current))?;`

By calling `create()` StateFactory comes into account.

contracts/collections/sg721-base/src/msg.rs
taitruong: This is for generating docs in JSON files. So we must keep this.

contracts/collections/sg721-base/src/contract.rs
taitruong: We need this. `T` is for generic. If it would be e.g. just `NftExtension` than it is a specific struct. Look how `Sg721Contract` is used:

```rs
        Sg721Contract::<
            DefaultOptionalNftExtension,
            DefaultOptionalNftExtensionMsg,
            DefaultOptionalCollectionExtension,
            DefaultOptionalCollectionExtensionMsg,
            Empty,
            Empty,
            Empty,
        >::default()
        .execute(deps, env, info, msg)
```

Now if Stargaze lateron wants to provide another custom contract sg721, and provide a different type like MyCustomNftExtension for `TNftExtension`. Then it may look like this:

```rs
        Sg721Contract::<
            MyCustomNftExtension,
            MyCustomNftExtensionMsg, // optional custom msg, but doesnt has to
            DefaultOptionalCollectionExtension,
            DefaultOptionalCollectionExtensionMsg,
            Empty,
            Empty,
            Empty,
        >::default()
        .execute(deps, env, info, msg)
```

taitruong: > RE naming conventions. I'm just curious if we need to name variables this way. The comment says NFTInfo extension msg for onchain metadata. Would ie be wrong to name it NFTInfoOnchainMetadataExtensionMsg or OnchainMetadataExtensionMsg something like that? I see this pattern in a lot of variables where the description is different than the name. Or do we need these names for consistency?

I should need change comment, since custom contracts can use this for any logic and type - not only for onchain metadata. This is also the reason why I used `extension` (instead of metadata) as naming for all structs, methods, generics, etc.

contracts/collections/sg721-base/src/contract.rs
```diff
@@ -328,88 +237,82 @@ where
...
-        }
-
-        #[allow(clippy::cmp_owned)]
-        if prev_contract_version.version < "3.1.0".to_string() {
-            response = crate::upgrades::v3_1_0::upgrade(deps.branch(), &env, response)?;
-        }
-
-        cw2::set_contract_version(deps.storage, CONTRACT_NAME, CONTRACT_VERSION)?;
+        // these upgrades can always be called. migration is only executed in case new stores are empty! It is safe calling these on any version.
+        response = crate::upgrades::v3_0_0_ownable_and_collection_info::upgrade(
```
taitruong, line 305: @MightOfOaks check this out for migration!

